PR #39, opened 2026-02-24
fix: update l34route backendref to use L34BackendRef

ljkiraly:


api/v1alpha1/l34route_types.go
```diff
@@ -130,6 +120,28 @@ type L34RouteSpec struct {
 	ByteMatches []string `json:"byteMatches,omitempty"`
 }
 
+// L34BackendRef defines a reference to a backend in the L34Route.
+type L34BackendRef struct {
+	// Group is the API group of the referent.
+	// +kubebuilder:default="gateway.networking.k8s.io"
```
zolug: Should default to the right "url" not the gateway one.

zolug:
Overall if we want L34Route to default to DistributionGroup as backendref I don't see any other way.
Service as a backend wouldn't make much sense for us as we are aiming for secondary networks. Maybe rejecting Service kind in the webhook would make sense.

Issue #37, opened 2026-02-24
LB controller logic: Nftables rules management for traffic steering


ljkiraly:
Handle nftables updates (add, delete, update). 
Clean up rules on shutdown

PR #51, opened 2026-03-06
LB controller logic: Nftables rules management for traffic steering

ljkiraly:
Add nftables VIP rules implementation

Issue: gh-37

Implement nftables manager for VIP traffic queuing:
- Create table per DistributionGroup
- IPv4/IPv6 VIP sets with interval support
- Prerouting chain to queue VIP traffic to nfqueue
- Dynamic VIP updates from L34Route destinationCIDRs

Integration with LoadBalancer controller:
- Extract VIPs from L34Routes
- Configure nftables before NFQLB flows
- Add nftablesManager interface for abstraction
- Add NftManagerFactory field to controller for testability
- Create nftables manager in reconcileNFQLBInstance
- Configure VIP rules in reconcileFlows before NFQLB flows
- Cleanup nftables on DistributionGroup deletion

Add readiness file for LoadBalancer instances

Testing:
- Add mockNftablesManager for unit tests

Flow:
1. DistributionGroup created → nftables manager created
2. L34Routes configured → VIPs extracted → nftables updated
3. NFQLB flows configured
4. DistributionGroup deleted → nftables cleaned up


internal/controller/loadbalancer/controller.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	// Check if route references this Gateway
+	if !c.referencesGateway(route) {
+		return nil
+	}
+
+	// Enqueue all DistributionGroups referenced by this route
+	var requests []ctrl.Request
+	for _, backendRef := range route.Spec.BackendRefs {
```
zolug: We are using Gateway API's BackendObjectReference. Must use defaulting rules according to its [specs](https://pkg.go.dev/sigs.k8s.io/gateway-api/apis/v1#BackendObjectReference).

docs/code-review-lb-controller.md
```diff
@@ -0,0 +1,485 @@
...
+**Commit b7d0158 - Refactoring**:
+- ✅ Split 543-line controller.go into 4 focused files:
+  - `controller.go` (215 lines) - Core reconciliation
+  - `instance.go` (72 lines) - NFQLB instance management
+  - `targets.go` (116 lines) - Target activation/deactivation
+  - `flows.go` (235 lines) - Flow configuration
+- ✅ **Rationale**: Improves maintainability without over-engineering
+
+**Commit 2ac8037 - nftables Integration**:
+- ✅ Table per DistributionGroup (isolation)
```
zolug: We do not want dedicated tables per DistributionGroups. That would not scale well, as it would lead to multiple copies of almost the same nftables rules. And packets would/could be subject to multiple re-injection into the same nfqueue process in case of overlapping VIPs (considered rather common).
These rules have been designed to be static and "shareable" and dynamically adjustable via the usage of nftables sets, where only the set content needs to change. (By concept we only expect to have a single nfqlb process, all packets destined to VIPs should be steered towards the same nfqueue target.)

I would say we could employ a separate second reconciliation loop, watching the parent GW of the LB only and extract the VIPs from `status.addresses` (the GW controller should expose VIPs via that field) to synchronize the nftables sets representing VIPs.

internal/controller/loadbalancer/instance.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if c.nftManagers == nil {
+		c.nftManagers = make(map[string]nftablesManager)
+	}
+	if c.targets == nil {
+		c.targets = make(map[string]map[int][]string)
+	}
+
+	// Check if instance already exists
+	if _, exists := c.instances[distGroup.Name]; exists {
```
zolug: (In the future, we might consider comparing existing DG params against incoming in case we should support maxEndpoint updates or such.)

internal/controller/loadbalancer/controller.go
```diff
@@ -0,0 +1,234 @@
...
+	for i := range l34routeList.Items {
+		route := &l34routeList.Items[i]
+
+		// Check if route references this Gateway
+		if !c.referencesGateway(route) {
+			continue
+		}
+
+		// Check if route references this DistributionGroup
+		for _, backendRef := range route.Spec.BackendRefs {
```
zolug: gatewayapiv1.BackendRef spec should be followed for defaulting etc

docs/code-review-lb-controller.md
ljkiraly: Resolved by commit bbec325 Refactor to single shared nftables table with Gateway VIPs

Issue #56, opened 2026-03-18
EndpointNetworkConfiguration Controller Logic


ljkiraly:
Implement the controller-manager logic that creates and reconciles `EndpointNetworkConfiguration` (ENC) resources for application Pods. The ENC declares the desired network state (VIPs, next-hops, network identity) that the sidecar controller (`internal/controller/sidecar/`, branch `app-sidecar`) applies to the Pod's network namespace.

This issue covers the **producer** side — the controller-manager logic that populates the ENC spec.

### Primary Resource: Pod

The controller reconciles **Pods** as the primary resource. For each Pod, it determines which Gateways the Pod participates in (via DistributionGroup → L34Route → Gateway chain), computes the desired network state, and creates/updates an ENC resource with the same name as the Pod.

### ENC Lifecycle

| Event | Action |
|-------|--------|
| Pod created + matches DG selector | Create ENC (same name as Pod) |
| Gateway.status.addresses changes | Update ENC VIPs |
| SLLBR Deployment scales | Update ENC next-hops |
| L34Route added/removed | Update ENC VIPs (via Gateway.status.addresses) |
| Pod deleted | ENC garbage-collected via ownerReference |
| Pod no longer matches any DG | Delete ENC |

### Watched Resources

| Resource | Watch Type | Trigger |
|----------|-----------|---------|
| Pod | `For` (primary) | Pod create/update/delete |
| DistributionGroup | `Watches` + mapper | DG selector or parentRef changes |
| Gateway | `Watches` + mapper | status.addresses changes |
| L34Route | `Watches` + mapper | parentRef/backendRef changes |
| GatewayConfiguration | `Watches` + mapper | networkSubnets changes |
| Deployment (SLLBR) | `Watches` + mapper | SLLBR Pod IPs change |

### MVP Scope

- [ ] Controller reconciling Pods → ENC create/update/delete
- [ ] Resolve Pod → DistributionGroup (via label selector matching)
- [ ] Resolve DistributionGroup → Gateway (via parentRefs + L34Route chain)
- [ ] Extract VIPs from `Gateway.status.addresses` (split by IP family)
- [ ] Extract next-hops from SLLBR Pod secondary IPs
- [ ] Extract network identity from `GatewayConfiguration.spec.networkSubnets`
- [ ] OwnerReference from ENC → Pod (GC on Pod deletion)
- [ ] Filter by Gateway Accepted condition (only process Gateways managed by this controller)
- [ ] Unit tests with fake client
[...]

PR #57, opened 2026-03-19
Add ENC controller

ljkiraly:
Related issue: #56 

Implement the ENC controller that runs inside the controller-manager
and creates/updates/deletes ENC custom resources for application Pods.

The controller resolves the chain Pod → DistributionGroup → L34Route →
Gateway → GatewayConfiguration to determine what network configuration
each Pod needs (VIPs, next-hops, network identity), and writes it to an
ENC resource consumed by the sidecar controller.

### DIFF
--- a/internal/controller/endpointnetworkconfiguration/controller.go
+++ b/internal/controller/endpointnetworkconfiguration/controller.go
@@ -134,13 +134,12 @@ func (r *Reconciler) reconcileENC(ctx context.Context, pod *corev1.Pod, connecti
 		},
 	}
 
-	if err := ctrl.SetControllerReference(pod, desired, r.Scheme); err != nil {
-		return fmt.Errorf("failed to set owner reference: %w", err)
-	}
-
 	var existing meridio2v1alpha1.EndpointNetworkConfiguration
 	err := r.Get(ctx, client.ObjectKeyFromObject(desired), &existing)
 	if apierrors.IsNotFound(err) {
+		if err := ctrl.SetControllerReference(pod, desired, r.Scheme); err != nil {
+			return fmt.Errorf("failed to set owner reference: %w", err)
+		}
 		return r.Create(ctx, desired)
 	}
 	if err != nil {


bszirtes (approved):
## Specific Code Issues

### 🔴 Critical Issues

**None** - All code is production-ready.

### 🟡 Medium Priority Issues

**1. Sidecar Restart Recovery Not Implemented**

**Files**: `internal/controller/sidecar/controller.go`, `internal/controller/sidecar/network.go`

**Problem**: On restart (Pod restart, OOM kill, node reboot), all in-memory state is lost (`tableIDs`, `managedVIPs`, `nl`), causing:
- VIP leaks (old VIPs not removed)
- EEXIST errors on VIP re-add
- Orphaned routes/rules when table IDs shift

**Recommendation**: Implement restart recovery:
1. Seed `managedVIPs` from kernel state on startup (list addresses, filter /32 and /128)
2. Make `AddrAdd` tolerate EEXIST errors:
```go
if err := nl.AddrAdd(link, addr); err != nil {
   if !errors.Is(err, syscall.EEXIST) {
       return managed, err
   }
   // EEXIST is OK - VIP already exists
}
managed[vip] = true
```
3. Sweep managed table ID range on startup, remove stale rules/routes

**Impact**: High - documented as known limitation, should be addressed before production use.

---

**2. Status Update Conflict Handling**

**File**: `internal/controller/sidecar/controller.go:267-286`

**Problem**: Conflict errors don't trigger requeue.

**Recommendation**:
```go
if err := c.Status().Update(ctx, enc); err != nil {
   if apierrors.IsConflict(err) {
       return err  // Trigger requeue
   }
   return err
}
```
**Impact**: Medium - status updates may be lost on concurrent modifications.

### 🟢 Low Priority Issues

**1. Missing Reconcile Logic in ENC Readme**

**File**: `internal/controller/endpointnetworkconfiguration/readme.md`

**Problem**: Readme explains resolution chain but not reconcile logic steps.

**Recommendation**: Add reconcile logic section:
markdown
## Reconcile Logic

1. Fetch the Pod. If not found → no-op. If not Running → delete stale ENC if exists.
2. Resolve all GatewayConnections via the chain above.
3. If no connections → delete ENC if exists (Pod no longer targeted).
4. If connections exist → create or update the ENC (skip update if spec unchanged).

---

**2. No Test Parallelization**

**File**: `internal/controller/distributiongroup/controller_test.go`

**Problem**: Tests run sequentially, slower than necessary.

**Recommendation**: Add `t.Parallel()` to independent tests:
```go
func TestReconcile_DGNotFound(t testing.T) {
   t.Parallel()
   // ... test logic
}
```
---

**3. Slice Pre-allocation Opportunities**

**File**: `internal/controller/endpointnetworkconfiguration/resolve.go`

**Lines**: 215, 230, 240, 302, 308

**Recommendation**: Pre-allocate slices when size is known:
```go
accepted := make([]gatewayv1.Gateway, 0, len(gatewayMap))
```
---

**4. Template Loading on Every Reconcile**

**File**: `internal/controller/gateway/validation.go`

**Problem**: Template loaded from disk on every reconcile.

**Recommendation**: Cache template in reconciler struct with sync.RWMutex.

---

**5. `podsForGatewayKeys` Complexity**

**File**: `internal/controller/endpointnetworkconfiguration/mappers.go:126`

**Problem**: 80+ line function doing 3 distinct operations.

**Recommendation**: Extract helper functions for readability.

---

**6. `SetupWithManager` Not Tested**

**File**: `internal/controller/endpointnetworkconfiguration/controller.go:101`

**Problem**: Watch setup has 0% coverage.

**Recommendation**: Add integration tests (envtest) to verify watch triggers.

internal/controller/endpointnetworkconfiguration/resolve.go
```diff
@@ -0,0 +1,454 @@
...
+	var gwConfig meridio2v1alpha1.GatewayConfiguration
+	if err := r.Get(ctx, client.ObjectKey{Namespace: gw.Namespace, Name: ref.Name}, &gwConfig); err != nil {
+		return subnetToType, subnetToHint, client.IgnoreNotFound(err)
+	}
+
+	// Build interface hint from NAD attachments
+	var nadInterface string
+	for _, att := range gwConfig.Spec.NetworkAttachments {
+		if att.Type == attachmentTypeNAD && att.NAD != nil && att.NAD.Interface != "" {
+			nadInterface = att.NAD.Interface
```
arp-est: 2. getNetworkContexts only uses the first NAD interface as hint — it breaks out of the loop after the first NAD attachment:

```go
   for _, att := range gwConfig.Spec.NetworkAttachments {
       if att.Type == attachmentTypeNAD && att.NAD != nil && att.NAD.Interface != "" {
           nadInterface = att.NAD.Interface
           break
       }
   }
```

  Then uses that single interface as the hint for all subnets. If you have multiple NADs on different subnets (which the GatewayConfiguration supports up to 10), the hint will be wrong for all but the first subnet's
interface. The sidecar falls back to scanning all interfaces when the hint doesn't match, so it still works — just slower.

Issue #59, opened 2026-03-19
Gateway Router


arp-est:
An issue to collect tasks related to gateway router after mvp

PR #79, opened 2026-03-27
Adjust the go template generating bird config

arp-est:
- By default it should use bfd off.
- Only interfaces between routers are now present in the bfd secion.
- Added some tests

Restricting the multihop didn't seem necessary, according to the bird config, and our 1000 handed ai companion for multihop setup you would need the multihop {} field set, so present should only be single hop.

https://github.com/Nordix/Meridio-2/issues/59?issue=Nordix%7CMeridio-2%7C72

### DIFF
--- a/internal/bird/bird.go
+++ b/internal/bird/bird.go
@@ -20,8 +20,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"os"
 	"os/exec"
+	"slices"
 	"strings"
 	"sync"
 	"syscall"
@@ -158,13 +160,26 @@ func (b *Bird) generateConfig(vips []string, routers []*meridio2v1alpha1.Gateway
 		}
 	}
 
+	slices.SortFunc(routers, func(a, b *meridio2v1alpha1.GatewayRouter) int {
+		if a.Name < b.Name {
+			return -1
+		}
+		if a.Name > b.Name {
+			return 1
+		}
+		return 0
+	})
+
+	ifset := make(map[string]bool)
 	for _, r := range routers {
 		rd, err := toRouterData(r)
 		if err != nil {
 			return "", err
 		}
 		data.Routers = append(data.Routers, rd)
+		ifset[r.Spec.Interface] = true
 	}
+	data.BGPInterfaces = slices.Sorted(maps.Keys(ifset))
 
 	var buf strings.Builder
 	if err := birdConfigTmpl.Execute(&buf, data); err != nil {
--- a/internal/bird/bird_test.go
+++ b/internal/bird/bird_test.go
@@ -137,7 +137,7 @@ template bgp BGP_TEMPLATE {
 	debug {events, states};
 	direct;
 	hold time 90;
-	bfd on;
+	bfd off;
 	graceful restart off;
 	setkey off;
 	ipv4 {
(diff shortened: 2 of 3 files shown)


ljkiraly (approved):
Seems good for me.

zolug:
> Restricting the multihop didn't seem necessary, according to the bird config, and our 1000 handed ai companion for multihop setup you would need the multihop {} field set, so present should only be single hop.

As discussed before, the goal was to restrict bird to not open the multi-hop bfd listener port. Which gets opened with the current config though. Pls double-check before accepting AI answers. And refer to [Meridio v1](https://github.com/Nordix/Meridio/pull/554)... 

```
udp         UNCONN       0            0                                 0.0.0.0:3784                            0.0.0.0:*           users:(("bird",pid=2304096,fd=11))
udp         UNCONN       0            0                                 0.0.0.0:4784                            0.0.0.0:*           users:(("bird",pid=2304096,fd=12))
udp         UNCONN       0            0                                    [::]:3784                               [::]:*           users:(("bird",pid=2304096,fd=13))                  
udp         UNCONN       0            0                                    [::]:4784                               [::]:*           users:(("bird",pid=2304096,fd=14))
```

internal/bird/bird.go
```diff
@@ -155,12 +155,17 @@ func (b *Bird) generateConfig(vips []string, routers []*meridio2v1alpha1.Gateway
 		}
 	}
 
+	ifset := make(map[string]bool) // unique interface set (preserves insertion order via slice)
```
zolug: I might not get the comment. But I don't see anything that guarantees two consecutive reconcile operation will result in the same interface order thus avoiding _unnecessary changes_. Of course there's no comparison against a previous config, but I wonder how bird would be affected if the order of let's say two interfaces were to change (AFAIK, the controller-runtime cache uses a map meaning ordering in list operations is undeterministic).

arp-est: I don't think my plan was that grandiose, I think I just wanted the order of the interfaces under bfd { ... }, to match the order of the protocol bgp entries, since in a regular set it would become ordered I think.
something like:

bfd {
if_B,
if_A,
}

protocol bgp NBR-B { interface if_B ... }
protocol bgp NBR-A { interface if_A ... }

though not sure whether it does that now that I'm looking at it again :'D

arp-est:
> > Restricting the multihop didn't seem necessary, according to the bird config, and our 1000 handed ai companion for multihop setup you would need the multihop {} field set, so present should only be single hop.
> 
> As discussed before, the goal was to restrict bird to not open the multi-hop bfd listener port. Which gets opened with the current config though. Pls double-check before accepting AI answers. And refer to [Meridio v1](https://github.com/Nordix/Meridio/pull/554)...
> 
> ```
> udp         UNCONN       0            0                                 0.0.0.0:3784                            0.0.0.0:*           users:(("bird",pid=2304096,fd=11))
> udp         UNCONN       0            0                                 0.0.0.0:4784                            0.0.0.0:*           users:(("bird",pid=2304096,fd=12))
> udp         UNCONN       0            0                                    [::]:3784                               [::]:*           users:(("bird",pid=2304096,fd=13))                  
> udp         UNCONN       0            0                                    [::]:4784                               [::]:*           users:(("bird",pid=2304096,fd=14))
> ```

My bad, I've skimmed through the docs, but I overlooked it, will fix.

szvincze (approved):
LGTM

PR #86, opened 2026-03-31
Add OCI image labels for build metadata

bszirtes:
## Summary

Adds OCI standard image labels to all four container images to expose git commit SHA and build timestamp as image metadata.

## Changes

- Added `org.opencontainers.image.revision` label (git commit SHA)
- Added `org.opencontainers.image.created` label (build timestamp)
- Applied to all images:
  - `controller-manager`
  - `network-sidecar`
  - `stateless-load-balancer`
  - `router`

## Implementation

The labels are added to the final stage of each Dockerfile using the existing `GIT_COMMIT` and `BUILD_DATE` build args that are already passed from the Makefile and embedded in the binaries.

## Benefits

- **Traceability**: Easily identify which git commit produced an image
- **Tooling integration**: Container registries (Docker Hub, GHCR) automatically display OCI labels
- **Consistency**: Same metadata in both binary (`--version`) and image labels
- **Standards compliance**: Follows [OCI image spec annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md)

### DIFF
--- a/build/controller-manager/Dockerfile
+++ b/build/controller-manager/Dockerfile
@@ -36,6 +36,8 @@ FROM alpine:3.22
 ARG USER
 ARG UID
 ARG HOME=/home/${USER}
+ARG GIT_COMMIT=unknown
+ARG BUILD_DATE=unknown
 
 RUN addgroup -g ${UID} ${USER} \
   && adduser -D -h ${HOME} -u ${UID} -G ${USER} ${USER} \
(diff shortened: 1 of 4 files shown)


ljkiraly (approved):

alitta-german (approved):
Thanks

zolug:
Duplicated defaults look a bit strange (GIT_COMMIT, BUILD_DATE), but still okay.
(In case they were global args the defaults would live in one place.)

PR #91, opened 2026-04-01
Expose birds logging related parameters

arp-est:
Made the logging realted params into a list,
You can add arbitrary log entries to bird this way, and each of them should work.

### DIFF
--- a/internal/bird/log.go
+++ b/internal/bird/log.go
@@ -1,0 +1,222 @@
+/*
+Copyright (c) 2026 OpenInfra Foundation Europe. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bird
+
+import (
+	"fmt"
+	"slices"
+	"strconv"
+	"strings"
+)
+
+// BirdLog represents a single BIRD log destination.
+// Maps to the BIRD config directive:
+//
+//	log "filename" [size "backup"] | fixed "filename" size | syslog [name name] | stderr | udp address [port port] all|{ list of classes }
+type BirdLog struct {
+	Type       string   // "file", "fixed", "syslog", "stderr", "udp"
+	Path       string   // file path, syslog name, or UDP address
+	Size       int      // rotation limit in bytes (file) or ring buffer size (fixed)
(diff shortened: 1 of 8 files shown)


internal/bird/log.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+		return BirdLog{Type: typ, Path: parts[1], Size: size, Classes: parseClasses(parts[3])}, nil
+
+	case "syslog":
+		if len(parts) != 3 {
+			return BirdLog{}, fmt.Errorf("invalid syslog log spec %q: expected syslog:name:classes", s)
+		}
+		return BirdLog{Type: typ, Path: parts[1], Classes: parseClasses(parts[2])}, nil
+
+	case "udp":
```
ljkiraly, line 120: udp:[::1]:514:all would be parsed incorrectly. Low prio since UDP logging to IPv6 is rare, but worth documenting as a known limitation

internal/common/config/router.go
```diff
@@ -79,6 +93,21 @@ func (c *RouterConfig) BindEnv(fs *pflag.FlagSet) {
...
+// bindBirdLogs binds a semicolon-separated environment variable to a BirdLogParams.
+// Each entry is parsed as a bird log spec (e.g. "stderr:all;file:/var/log/bird.log:all").
+// Only applies if the corresponding flag was not explicitly set.
+func bindBirdLogs(fs *pflag.FlagSet, flagName, envName string, target *bird.BirdLogParams) {
+	if !fs.Changed(flagName) {
+		if val := os.Getenv(envName); val != "" {
+			for entry := range strings.SplitSeq(val, ";") {
+				entry = strings.TrimSpace(entry)
+				if entry != "" {
+					_ = target.Set(entry)
```
ljkiraly, line 111: Consider adding a default log config if we can have a silent Bird :)
If no '--bird-log' flag is set and 'MERIDIO_BIRD_LOG' is empty the BIRD default (no logging) will be set. Previously 'stderr:all' was set - not sure if that is better.

ljkiraly (commented):
In general looks good. After a rebase, I can approve.
+ add license headers to new files

internal/bird/log.go
arp-est: Added handling for ipv6 address, it now has the option to put ipv6 addresses inside [], and that will be used as is regardless if it contains separators.
for example udp:[::1]:543:all , will interpret everything inside [] as the address, its only for ipv6.

internal/common/config/router.go
arp-est: I am not sure either :'D, I mean I don't have anything against not having logging configured.
According to the bird docs, by default bird logs everything to syslog.

I've fiddled with it a bit, but seems to be a bit cumbersome to set a default nicely in cobra

ljkiraly (approved):

arp-est:
Just to mention it here too. So checked the default bird logging, it does log to syslog, and that by default puts the logs to /var/log/messages. If no syslogd is running, then it does not put logs there.

Issue #60, opened 2026-03-19
Test netlink commands in setPolicyRoutes


arp-est:
There should be a mock interface to check whether the netlink commands are done correctly.

PR #96, opened 2026-04-08
Add routingOps interface for testable policy routing in router

szvincze:
This PR solves issue #60.

Extract netlink calls from setPolicyRoutes and setupBlackholeRoutes into a routingOps interface, following the same pattern as the sidecar controller's netlinkOps.

- Introduce a routingOps interface (RuleListFiltered, RuleAdd, RuleDel, RouteReplace) with a defaultRoutingOps implementation that delegates to real netlink
- Add an nl field to the Bird struct, initialized in New()
- Add 13 unit tests covering:
  - add/remove/idempotent VIP rules
  - dual-stack scenarios
  - mixed add/remove operations
  - best-effort error accumulation
  - blackhole route setup
  - error injection for all netlink operations


internal/bird/netlink.go
```diff
@@ -0,0 +1,37 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bird
+
+import "github.com/vishvananda/netlink"
+
+// routingOps abstracts netlink syscalls used by policy routing for testability.
+type routingOps interface {
```
arp-est: I don't like the name, it fiddles with rules, and routes, and its just an interface for some netlink calls, to be able to test it. I would have probably called it AbstractNetlink or something along those lines.
And then the defaultRoutingOps could be called netlink, and the mock, netlinkMock or something.

In fact can't you just pass netlink itself without implementing defaultRoutingOps?

Checked with AI, there is a &netlink.Handle{} thing that can be used as a direct replacement for the defaultRoutingOps, since the method signatures are the same.

szvincze: Please see 649a33e where I added the requested changes.

arp-est: looks good to me :+1:

arp-est (approved):

PR #97, opened 2026-04-09
Stop router if either bird, or monitor fails.

arp-est:
https://github.com/Nordix/Meridio-2/issues/62
https://github.com/Nordix/Meridio-2/issues/61

It turned out to be pretty basic, just added a cancel, to the context used for the monitor, bird, and the manager.

I'm not quite sure whether we want to fail in cases if for example birdc call fails, but I'm guessing its best to leave it for the time when we do the VIP exposure on the readiness gates

### DIFF
--- a/cmd/router/cmd/run.go
+++ b/cmd/router/cmd/run.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/sync/errgroup"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -151,18 +152,28 @@ func runRouter(ctx context.Context, cfg *config.RouterConfig) error {
 	setupLog.Info("starting router", "gateway", cfg.GatewayName, "namespace", cfg.GatewayNamespace)
 
 	ctx = ctrl.SetupSignalHandler()
+	g, ctx := errgroup.WithContext(ctx)
 
-	go func() {
+	g.Go(func() error {
 		if err := birdInstance.Run(ctx); err != nil {
-			setupLog.Error(err, "BIRD stopped")
+			return fmt.Errorf("BIRD stopped: %w", err)
 		}
-	}()
+		if ctx.Err() == nil {
+			return fmt.Errorf("BIRD exited unexpectedly")
+		}
+		return nil
+	})
 
-	// Start monitoring BGP connectivity
-	go monitorConnectivity(ctx, mgr, birdInstance)
+	g.Go(func() error {
+		return monitorConnectivity(ctx, mgr, birdInstance)
+	})
 
-	if err := mgr.Start(ctx); err != nil {
-		setupLog.Error(err, "problem running manager")
+	g.Go(func() error {
+		return mgr.Start(ctx)
+	})
+
+	if err := g.Wait(); err != nil {
+		setupLog.Error(err, "router exited")
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/router/cmd/run.go
```diff
@@ -148,15 +148,17 @@ func runRouter(ctx context.Context, cfg *config.RouterConfig) error {
...
 	go func() {
 		if err := birdInstance.Run(ctx); err != nil {
 			setupLog.Error(err, "BIRD stopped")
 		}
+		cancel()
 	}()
 
 	// Start monitoring BGP connectivity
-	go monitorConnectivity(ctx, mgr, birdInstance)
+	go monitorConnectivity(ctx, cancel, mgr, birdInstance)
```
zolug: It's generally discouraged in Go to pass around context cancel func as parameter. The concern is about ownership, i.e. whoever creates the cancellable context should own the cancel. Passing it around makes it unclear who's responsible for that.
AFAIK the idiomatic Go approach is using err channel, but feel free to double check. 
In general, enforce idiomatic, production grade Go solutions.

arp-est: I couldn't find any mention about the channels, I'm suspecting you mean, when a routing is actively polling a channel, and reacting to ctx.Done(), which is for reacting to a cancel, but not initiating it.

But, I did find this [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) go package, that seems to be a more idiomatic way to run goroutines.

It says, that if one of the goroutines in the errgroup returns with nonzero exit, it will cancel all members of the group.

cmd/router/cmd/run.go
```diff
@@ -148,39 +149,44 @@ func runRouter(ctx context.Context, cfg *config.RouterConfig) error {
...
+	g, ctx := errgroup.WithContext(ctx)
 
-	go func() {
+	g.Go(func() error {
 		if err := birdInstance.Run(ctx); err != nil {
-			setupLog.Error(err, "BIRD stopped")
+			return fmt.Errorf("BIRD stopped: %w", err)
 		}
-	}()
+		return fmt.Errorf("BIRD exited unexpectedly")
```
zolug: Wouldn't this return "BIRD exited unexpectedly" error even on graceful shutdown?

wouldn't the following be more accurate?

```go
g.Go(func() error {
    if err := birdInstance.Run(ctx); err != nil {
        return fmt.Errorf("BIRD stopped: %w", err)
    }
    if ctx.Err() == nil {
        return fmt.Errorf("BIRD exited unexpectedly")
    }
    return nil
})
```

- BIRD crashes (returns error) -> propagate error (OK)
- BIRD exits cleanly but context still active (shouldn't happen, defensive) -> error (OK)
- BIRD exits cleanly because context cancelled by signal (graceful shutdown) -> nil (OK)
- BIRD exits cleanly because context cancelled by errgroup (another member failed) -> nil (OK)

cmd/router/cmd/run.go
```diff
@@ -189,10 +195,10 @@ func monitorConnectivity(ctx context.Context, mgr ctrl.Manager, birdInstance *bi
 	for {
 		select {
 		case <-ctx.Done():
-			return
+			return nil
 		case status, ok := <-statusCh:
 			if !ok {
-				return
+				return fmt.Errorf("monitor channel closed unexpectedly")
```
zolug, line 210: When ctx.Done() and statusCh are ready simultaneously, select picks randomly. If ctx.Done() wins, return nil. If statusCh wins, it returns an error.

There can be a race between birdInstance.Monitor's internal goroutine and monitorConnectivity's select during graceful shutdown IMHO, that could cause spurious errors.

In theory, in birdInstance.Monitor there could other exit paths other than when ctx.Done() fires, but currently that's not the case. So, statusCh closing always means the context was cancelled. It cannot close "unexpectedly" in the current implementation.

Not a severe issue or anything. But just in case we could do:

```go
case status, ok := <-statusCh:
    if !ok {
        if ctx.Err() != nil {
            return nil
        }
        return fmt.Errorf("monitor channel closed unexpectedly")
    }

```

However, if Monitor() ever gains an error path that closes the channel, the consumer can't reliably distinguish it from context cancellation due to the select race. Hence, it's recommended that the error is logged inside Monitor() itself before closing the channel (the consumer just sees "channel closed" either way).

arp-est: done

cmd/router/cmd/run.go
arp-est: done

zolug (approved):

Issue #85, opened 2026-03-31
LB controller; PMTU handling


ljkiraly:
Set nftables rules and sysctls required to handle MTU differences between external and internal networks.

### What It Does
Rewrites source address of ICMP "Fragmentation Needed" / "Packet Too Big" replies to use the VIP address, ensuring external clients can receive PMTU discovery messages.

### Problem It Solves
When a packet destined to a VIP is too large for the internal network MTU:
1. Kernel generates ICMP error with **load balancer's IP** as source
2. External client receives ICMP from wrong IP → ignores it
3. PMTU discovery fails → connection hangs or performs poorly

**Solution**: Rewrite ICMP source to use the **VIP** from the original packet.

**Sysctl requirements**:
```bash
sysctl -w net.ipv4.fwmark_reflect=1
sysctl -w net.ipv6.fwmark_reflect=1
```

PR #99, opened 2026-04-09
Pmtu discovery

ljkiraly:
Related issue: gh-85

Add PMTU SNAT chain (snat-local) that rewrites source address of locally
generated ICMP Frag Needed / ICMPv6 Packet Too Big replies to use the VIP
from the encapsulated original packet. Add selective defrag table
(meridio-defrag) that enables kernel defragmentation for external traffic
while skipping it for configurable target-facing interface prefixes.

- PMTU SNAT: route-type output chain, fwmark-based early skip, payload
  write with checksum update (IPv4) and mark reset

> DEFRAG not needed
> - Defrag: pre-defrag chain at priority -500 (before conntrack at -400),
>   notrack per excluded interface prefix, dummy conntrack for defrag loading
> - Config: --defrag-excluded-if-prefixes flag / MERIDIO_DEFRAG_EXCLUDED_IF_PREFIXES
>   env var for target-facing interface names (this static flag was replaced by MERIDIO_NETWORK_SUBNET_CIDRS env var and auto interface discovery was implemented in LB controller)

- Sysctls: fwmark_reflect=1 added to sysctl-tuning NADs (required for
  PMTU SNAT mark check)
- Unit tests for PMTU expression structure (offsets, types, checksums)
- Integration tests with network namespace isolation (CAP_NET_ADMIN)
- E2e large-packet ping test for PMTU path validation


ljkiraly:
Defrag handling:  Meridio-2 uses Multus NADs (not NSM as Meridio), so the interface prefix for the `notrack` rule differs. The target-facing interface names come from NAD configuration rather than NSM naming. Need to determine the correct interface name(s) to exclude from 'defrag'. The 'defrag' feature is about not defragmenting packets coming FROM target-facing interfaces.
The problem: The LB pod doesn't currently discover its own interface topology at runtime. A CLI flag (MERIDIO_DEFRAG_EXCLUDED_IF_PREFIXES) is used to specify the excluded interfaces. The Gateway controller can inject  in the future by deriving app-network interface names from the GatewayConfiguration. 

- networkAttachments lists ALL NADs (vlan-100, app-net-a1, etc.) with their interface names
- networkSubnets lists the app-facing CIDRs with attachment type

We need a way to associate networkSubnets with specific NAD interfaces. But the current API doesn't have this mapping.

ljkiraly:
Turned out that defrag not needed at all since there is no connection tracking and no nftables rules which implies fragmentation/defragmentation of packets from target side.

docs/controllers/gateway.md
```diff
@@ -462,6 +462,70 @@ networkAttachments:
...
+
+**Required sysctls:**
+
+| Sysctl | Value | Purpose |
+|--------|-------|---------|
+| `net.ipv4.conf.all.forwarding` | `1` | IP forwarding (LB must forward packets) |
+| `net.ipv6.conf.all.forwarding` | `1` | IPv6 forwarding |
+| `net.ipv4.fib_multipath_hash_policy` | `1` | L4 (5-tuple) ECMP hashing |
+| `net.ipv6.fib_multipath_hash_policy` | `1` | L4 ECMP hashing for IPv6 |
+| `net.ipv4.conf.all.rp_filter` | `0` | Disable reverse path filtering (VIP traffic would fail RPF) |
```
zolug: must support running with rp_filter=2 (the primary network's pod default route ensures it shall work)

docs/controllers/gateway.md
```diff
@@ -462,6 +462,70 @@ networkAttachments:
...
+| `net.ipv4.conf.default.rp_filter` | `0` | Disable RPF for new interfaces |
+| `net.ipv6.conf.all.accept_dad` | `0` | Disable DAD (avoids delays on interface up) |
```
zolug: accept_dad - INTEGER
Whether to accept DAD (Duplicate Address Detection).

| 0 | Disable DAD
| 1 | Enable DAD (default)
| 2 | Enable DAD, and disable IPv6 operation if MAC-based duplicate link-local address has been found.

DAD operation and mode on a given interface will be selected according to the maximum value of conf/{all,interface}/accept_dad.

internal/nftables/manager.go
```diff
@@ -216,6 +222,134 @@ func (m *Manager) createOutputChain() error {
...
+	})
+
+	return m.conn.Flush()
+}
+
+// buildPMTUDIPv4Exprs builds nftables expressions for IPv4 PMTU SNAT.
+// Matches: ICMP type 3 (Dest Unreachable) code 4 (Frag Needed), mark != 0,
+// dst NOT VIP, src NOT VIP, encapsulated dst IS VIP.
+// Action: rewrite IP src to encapsulated dst (the VIP), reset mark to 0.
+func buildPMTUDIPv4Exprs(ipv4Set *nftables.Set) []expr.Any {
```
zolug, line 276: I had to look up the details from Meridio v1. Would be nice to "detail" the rational a bit more.

test/e2e/e2e_suite_appnetwork_test.go
```diff
@@ -120,6 +120,17 @@ var _ = Describe("E2E Test Suites", func() {
 					}
 				})
 
+				Context("ICMP large packet", func() {
+					for _, gw := range suite.gateways {
+						gw := gw
+						It("handles 1400-byte ping on "+gw.name+" VIP", func() {
+							// 1400 payload + 28 headers = 1428, fits within 1500 MTU.
+							// Exercises the output chain ICMP → nfqueue path with larger packets.
```
zolug: It rather takes "the primary path" since there's no MTU drop when packets reach the internal network. I modified the mtu in the internal NAD to test these SNAT based rules.

docs/controllers/gateway.md
ljkiraly: Fixed by bc269af. (b99c6ac)
Verified by e2e tests.

internal/nftables/manager.go
ljkiraly: Fixed by bc269af. (b99c6ac)

docs/controllers/gateway.md
ljkiraly: Fixed by bc269af. (b99c6ac)
Verified by e2e tests.

test/e2e/e2e_suite_appnetwork_test.go
ljkiraly: Fixed by 0abfcfb

test/e2e/utils/traffic.go
```diff
@@ -51,6 +52,101 @@ func Ping(vip string) error {
...
+
+	// Flush PMTU cache so the oversized packet is actually sent on the wire
+	// (otherwise the kernel rejects it locally from a previous PMTU discovery)
+	flushCmd := fmt.Sprintf("docker exec vpn-gateway ip route flush cache %s", vip)
+	_ = exec.Command("/bin/sh", "-c", flushCmd).Run()
+
+	// Build tcpdump filter for ICMP unreachable (type 3) from the VIP
+	var tcpdumpFilter string
+	if isIPv6 {
+		tcpdumpFilter = fmt.Sprintf("icmp6 and src host %s", vip)
```
zolug: Does not filter for packet-too-big

zolug (approved):

Issue #76, opened 2026-03-27
Transition controller-manager RBAC from ClusterRole to namespace-scoped Roles


zolug:
### Description:

The controller-manager currently uses a single ClusterRole (generated by kubebuilder RBAC markers) for all permissions. This grants cluster-wide access to namespace-scoped resources, 
violating least-privilege when the operator is deployed with --namespace scoping.

Meridio v1 used hand-crafted namespace-scoped Roles per component (operator, NSP, frontend), with ClusterRoles only for cluster-scoped resources. Meridio-2 should follow the same 
pattern.

### Proposed changes:

- Split the generated ClusterRole into:
  - A minimal ClusterRole for cluster-scoped resources only: GatewayClass (get/list/watch)
  - A namespace-scoped Role for everything else: Pods, Deployments, EndpointSlices, Gateways, L34Routes, GatewayConfigurations, DistributionGroups, EndpointNetworkConfigurations, etc.
- Hand-write the Role (like the existing config/rbac/lb-serviceaccount.yaml for LB pods) instead of relying on kubebuilder marker generation
- Update RoleBinding to bind the Role to the controller-manager ServiceAccount in the deployment namespace

PR #102, opened 2026-04-13
rbac: split ClusterRole into namespace-scoped Role and ClusterRole

zolug:
https://github.com/Nordix/Meridio-2/issues/76

Replace the generated ClusterRole with hand-maintained RBAC:
- ClusterRole for cluster-scoped resources only (GatewayClass, Nodes)
- Namespace-scoped Role for all other resources
- Remove kubebuilder RBAC markers from controller source files
- Stop generating RBAC via controller-gen


ljkiraly (commented):
No functional issues found. The split seems correct. The AI found some remaining referrences:

1. AGENTS.md has 3 stale references (not updated by the PR):
   - Line 13: config/rbac/role.yaml          Generated RBAC (DO NOT EDIT) — file no longer exists
   - Line 41: - config/rbac/role.yaml - from make manifests — no longer generated
   - Line 63: make manifests  # Regenerate CRDs/RBAC from markers — RBAC is no longer regenerated

  These should be updated to reflect that RBAC is now hand-maintained in manager-role.yaml and manager-clusterrole.yaml.

2. docs/operations/constraints-and-limitations.md item #2 is now resolved (not updated by the PR):
  > "The controller-manager's RBAC is a single ClusterRole granting cluster-wide access to namespace-scoped resources. Should be split into a namespace-scoped Role..."
  
  This is exactly what the PR does. The item should be marked as resolved (strikethrough) or removed.

3. docs/controllers/sidecar.md line 26 has a stale reference:
  > "No kubebuilder RBAC markers (would pollute shared role.yaml via make manifests)"
  
  Since role.yaml is no longer generated and markers are removed, this rationale is outdated. Should be updated to say the sidecar uses a separate ServiceAccount with its own RBAC (the reason is still valid — just the mechanism description is stale).

Issue #83, opened 2026-03-31
LB controller; remove nspAPI (Meridio) dependency and create internal NFQLB API implementation.


ljkiraly:
Create interfaces;
Update controller;
Update tests;
Validation, cleanup.

PR #109, opened 2026-04-16
LB controller; remove nspAPI (Meridio) dependency and create internal NFQLB API implementation

ljkiraly:
Related issue: gh-83

- Copy pkg/nfqlb from meridio-2.x POC as internal package.
- Adapt to MVP
- Replace github.com/nordix/meridio dependency with internal nfqlb package


ljkiraly:
Wait until fix for #129 is merged to master and apply the fix for the refactored code - same pattern: replace RouteAdd with RouteReplace, tolerate EEXIST from RuleAdd, remove validPolicyRoute.

zolug:
here's a preliminary rewiew:

```md
### Overall Assessment

Good refactor. Removes the heavy github.com/nordix/meridio dependency (NSM, OpenTelemetry, logrus) and replaces it with a focused internal/nfqlb package. Routing is now co-located with
target lifecycle. The nfqlb package API natively supports dual-stack, though the LB controller's EndpointSlice processing doesn't yet leverage this (#144).

### Dual-Stack Support

nfqlb package (API level): ✅ Supports dual-stack.

AddTarget(ctx, ips []string, identifier int) creates a policy route for each IP. createPolicyRoute determines the family from the IP, creates a family-specific rule + route. IPv4 and 
IPv6 coexist under the same fwmark (Linux ip rules are per-family, routing tables are independent per family within the same table ID).

LB controller (EndpointSlice processing): ❌ Not yet dual-stack.

targets.go still uses newTargets[identifier] = endpoint.Addresses (overwrite). Only the last-processed family survives. Pre-existing limitation, tracked by #144. The simplest fix 
aligned with this PR's direction: append(newTargets[identifier], endpoint.Addresses...).

### Routing (internal/nfqlb/routing.go)

- ✅ Uses RouteReplace — handles stale routes atomically
- ✅ ensureRule — prevents duplicate rule accumulation
- ⚠️ cleanNeighbor — called unconditionally on every createPolicyRoute, does not consider #138. See Heal Goroutine section for amplified impact.
- ⚠️ deletePolicyRoute returns error on RuleDel/RouteDel failure including "not found" (ENOENT/ESRCH). Currently harmless since reconcileTargets does not propagate errors. However, if 
the error-return requeue pattern (PR #139) is applied to this refactored code in the future, "not found" errors would cause infinite requeue loops: reconcile retries DeleteTarget for 
an already-clean target → gets ENOENT → requeues → repeats. At that point, deletePolicyRoute must ignore "not found" errors for idempotency.

### Offset Management (offset.go)

✅ Dynamic offset allocation — finds non-overlapping ranges based on actual maxTargets. Removes the old 1024-endpoint-per-DG hard limit.

⚠️ Doc update needed: docs/operations/constraints-and-limitations.md still documents the old formula and 1024 limit.

### Heal Goroutine

The heal goroutine periodically (every 10s) re-applies policy routes for all targets via doCreatePolicyRoute, which calls createPolicyRoute → cleanNeighbor unconditionally.

The heal is necessary given this design. AddTarget short-circuits when the target is already tracked with the same IPs:

go
if slicesEqual(existingIPs, ips) {
    return nil // doesn't verify routes exist in kernel
}


A reconcile-triggered retry would NOT fix a missing route — it short-circuits because in-memory state says "already configured." The heal is the only mechanism that re-applies routes for existing targets whose kernel state drifted.

Concern: The unconditional cleanNeighbor call on every heal tick (10s) flushes ARP/NDP entries for all stable targets, causing unnecessary traffic micro-interruptions (#138). Unlike 
Meridio v1 with NSM (dynamic interface changes at runtime), Meridio-2 uses Multus thin plugin with static interfaces — route drift without an API event is unlikely in practice.

Possible improvements:
- Implement #138 (skip neighbor flush for unchanged targets) — eliminates the traffic disruption while keeping the heal's route verification
- Alternatively, remove the short-circuit in AddTarget and rely on reconcile-based recovery (with error-return requeue + SyncPeriod ~1 hour as safety net), eliminating the need for the heal goroutine entirely

### Interface Design (nfqlb_iface.go)

✅ Clean nfqlbManager / nfqlbInstance interfaces for testability. NFQLBManagerAdapter wraps the concrete type.

### Error Return Pattern (PR #139 compatibility)

This PR's reconcileTargets returns nil unconditionally — does not include PR #139's errFinal error-return pattern. Merge order matters: if #109 merges after #139, the error-return 
pattern is lost.

Note: even with error-return requeue, the heal goroutine remains necessary because AddTarget short-circuits for existing targets (see above). The two mechanisms are complementary: 
error-return handles failures during state changes, heal handles drift for stable targets.

### Documentation Impact

docs/operations/constraints-and-limitations.md needs update: the old formula (DG_ID × 1024 + 5000) and 1024-endpoint hard limit are no longer accurate.

### Summary

| Aspect | Status |
|--------|--------|
| Dependency removal | ✅ Clean |
| Dual-stack (nfqlb API) | ✅ Handles multiple IPs per target |
| Dual-stack (LB controller) | ❌ Pre-existing overwrite, tracked by #144 |
| Routing correctness | ✅ RouteReplace + ensureRule |
| Self-healing | ✅ Necessary (AddTarget short-circuits existing targets), but causes unnecessary neighbor flushes every 10s (#138) |
| Testability | ✅ Interface-based |
| Offset management | ✅ Dynamic, 1024 limit removed |
| deletePolicyRoute idempotency | ⚠️ Not a problem now, becomes one if error-return requeue is added later |
| PR #139 compatibility | ⚠️ Error-return pattern not included, merge order matters |
| Documentation | ⚠️ Constraints doc needs update |

Observations to address before merge:
1. Heal goroutine and its unconditional neighbor flush (#138): document the trade-off or reconsider necessity given static network model
2. Update constraints doc for removed 1024 limit
3. Coordinate with PR #139 on error-return pattern

```

bszirtes (approved):
### 🔴 Critical Issues
**None** - All critical routing and error handling issues have been resolved.

---

### 🟡 Medium Priority Issues

#### 1. No upper bound check on offset allocation
**Location:** `internal/nfqlb/offset.go:30-35`

**Current code:**
```go
if offset >= (math.MaxInt - maxTarget + 1) {
    return 0, errIdentifierOffset
}
```

**Issue:** The check uses `math.MaxInt` which is extremely large. In practice, fwmark collisions could occur much earlier. With `startingOffset=5000` and `maxTarget=100`, the 1000th DistributionGroup would use fwmarks starting at ~105000, which is reasonable, but there's no practical limit enforced.

**Recommendation:**
```go
const maxDGCount = 1000  // Reasonable limit: 1000 DGs * 1024 = ~1M fwmark range

func getOffset(startingOffset int, instances map[string]*Instance, maxTarget int) (int, error) {
    offset := startingOffset
    dgCount := len(instances)
    
    if dgCount >= maxDGCount {
        return 0, fmt.Errorf("DG count limit reached (%d DGs)", maxDGCount)
    }
    
    // ... existing search logic
}
```

**Impact:** Low - unlikely to hit in practice, but would cause hard-to-debug fwmark collisions if it did.

---

#### 2. Hardcoded constants should be configurable
**Location:** `internal/nfqlb/const.go:30-33`

**Current code:**
```go
const (
    defaultQueue          = "0:3"
    defaultQLength        = 1024
    defaultStartingOffset = 5000
    defaultMaxTargets     = 100
)
```

**Issue:** These constants are hardcoded and cannot be adjusted for different deployment scenarios without code changes.

**Recommendation:** Add CLI flags to the loadbalancer controller:
```go
// In cmd/stateless-load-balancer/cmd/run.go
--nfqlb-queue string          NFQLB queue range (default "0:3")
--nfqlb-queue-length int      NFQLB queue length (default 1024)
--nfqlb-starting-offset int   Starting fwmark offset (default 5000)
```

**Impact:** Medium - limits flexibility in production deployments with different requirements.

---

#### 3. No validation that NFQLB process is running
**Location:** `internal/nfqlb/nfqlb.go:228-240` (AddInstance)

**Current code:**
```go
func (nfqlb *NFQueueLoadBalancer) AddInstance(ctx context.Context, name string, ...) (*Instance, error) {
    // No check if Start() was called or if process is running
    nfqlb.mu.Lock()
    defer nfqlb.mu.Unlock()
    // ... calls nfqlb CLI commands
}
```

**Issue:** If `Start()` hasn't been called or the process crashed, CLI commands will fail with cryptic errors like "connection refused" or "no such file".

**Recommendation:**
```go
type NFQueueLoadBalancer struct {
    // ...
    running atomic.Bool
}

func (nfqlb *NFQueueLoadBalancer) Start(ctx context.Context) error {
    nfqlb.running.Store(true)
    defer nfqlb.running.Store(false)
    // ... existing code
}

func (nfqlb *NFQueueLoadBalancer) AddInstance(...) (*Instance, error) {
    if !nfqlb.running.Load() {
        return nil, fmt.Errorf("NFQLB process not running (Start() not called or process crashed)")
    }
    // ... existing code
}
```

**Impact:** Medium - would provide clearer error messages for operational issues.

---

### 🟢 Low Priority Issues

#### 4. Silent error ignoring in IP change path
**Location:** `internal/nfqlb/nfqlb.go:478-479`

**Current code:**
```go
for _, ip := range existingIPs {
    _ = cleanNeighbor(net.ParseIP(ip))
    _ = s.doDeletePolicyRoute(fwmark, ip)
}
```

**Issue:** Errors from cleanup operations are silently ignored. While this is intentional (old routes might not exist), it could hide real issues.

**Recommendation:**
```go
for _, ip := range existingIPs {
    if err := cleanNeighbor(net.ParseIP(ip)); err != nil {
        ctrl.LoggerFrom(ctx).V(1).Info("Failed to clean neighbor (may not exist)", 
            "ip", ip, "error", err)
    }
    if err := s.doDeletePolicyRoute(fwmark, ip); err != nil {
        ctrl.LoggerFrom(ctx).V(1).Info("Failed to delete old route (may not exist)", 
            "ip", ip, "error", err)
    }
}
```

**Impact:** Low - mostly for debugging, doesn't affect functionality.

---

#### 5. Protocol validation error message could be clearer
**Location:** `internal/nfqlb/validate.go:91-95`

**Current code:**
```go
func validateProtocols(protocols []string) error {
    for _, p := range protocols {
        switch strings.ToLower(p) {
        case "tcp", "udp", "sctp":
            // valid
        default:
            return fmt.Errorf("unsupported protocol %q (allowed: tcp, udp, sctp)", p)
        }
    }
    return nil
}
```

**Issue:** Error message shows lowercase protocols but input might be uppercase (e.g., "TCP"), which could confuse users.

**Recommendation:**
```go
return fmt.Errorf("unsupported protocol %q (allowed: TCP, UDP, SCTP, case-insensitive)", p)
```

**Impact:** Low - minor UX improvement.

---

#### 6. Missing integration tests
**Location:** Test coverage gap

**Issue:** No integration tests for:
- Controller + nfqlb interaction (envtest-based)
- NFQLB process lifecycle (Start, context cancellation)
- Policy routing operations (netlink rule/route creation)
- RouteReplace and ensureRule behavior under concurrent reconciles

**Recommendation:** Add integration tests in a follow-up PR:
```go
// internal/controller/loadbalancer/integration_test.go
var _ = Describe("LoadBalancer Controller Integration", func() {
    It("should reconcile targets with real nfqlb package", func() {
        // Test with mock netlink but real nfqlb logic
    })
})
```

**Impact:** Low - unit tests provide good coverage, but integration tests would increase confidence.

zolug:
1. Partial failure handling in multi-step DeleteTarget

AddTarget and DeleteTarget are multi-step operations (nfqlb CLI activation/deactivation + route/rule creation/deletion). Partial failures leave the system in an 
intermediate state.

There are two levels of target tracking that should mirror each other but are updated independently:
- c.targets[dgName][identifier] (Controller level) — used for diffing old vs new in reconcileTargets, bulk-committed at end of reconcile
- s.targets[identifier] (nfqlb Instance level) — used inside AddTarget/DeleteTarget to determine code path, updated per-operation

AddTarget partial failures self-heal in all cases: if activate fails, s.targets is not updated → retry re-attempts activation. If activate succeeds but route creation 
fails, s.targets is committed → on requeue the idempotent re-apply path fixes routes. If the target disappears from EndpointSlices between retries, c.targets (
committed unconditionally) ensures DeleteTarget is called to clean up.

DeleteTarget partial failures do not self-heal: delete(s.targets, identifier) executes before doExec("deactivate"). If deactivation fails:
- s.targets no longer has the identifier (already deleted)
- c.targets[DG] = newTargets is committed unconditionally (also doesn't have it)
- Both maps agree "it's gone" — but nfqlb still has the target active in the hash table
- On requeue, the identifier appears in neither currentTargets nor newTargets → nobody calls DeleteTarget again → stuck

The semantics of an identifier being present in s.targets are unclear — does it mean "fully configured" or just "known"? This ambiguity makes it hard to reason about 
recovery from partial states. Fixing is non-trivial: deferring c.targets commit until success doesn't work cleanly because EndpointSlice content may change between 
retries, making the old c.targets stale for diffing.

The probability is low (nfqlb CLI failure for a running process is rare), and the impact is limited (one hash table slot with no route → packet dropped for that bucket
). But it's a correctness gap worth acknowledging.

Question: Is this considered acceptable risk, or should there be a mechanism to ensure failed deactivations are retried?

2. slicesEqual in AddTarget is order-sensitive

AddTarget uses slicesEqual(existingIPs, ips) to detect whether IPs changed. This comparison is order-dependent. If the caller builds the IP list from EndpointSlices 
whose processing order varies between reconciles (controller-runtime cache doesn't guarantee list order), the same set of IPs in different order triggers the "IPs 
changed" path — unnecessarily deleting routes, flushing neighbors, and re-creating routes on every reconcile.

Currently with single-family EndpointSlices (one IP per identifier) this isn't triggered, but with future dual-stack support (#144, append approach) it becomes a real 
problem.

Fix: normalize IP order before storing/comparing (e.g., sort), or use set-based comparison.

3. deletePolicyRoute is fragile

Two issues:
- **Bails out on RuleDel error** — if RuleDel fails (even "not found"), RouteDel is never attempted. The route is left behind.
- **Returns "not found" as error** — the desired state (rule/route gone) is achieved, yet the function reports failure.

Since all callers already ignore the error (_ = s.doDeletePolicyRoute(...)), the function should be made robust: attempt both deletions regardless of individual 
failures, ignore "not found" errors. Only return real errors (permission denied, etc.) if callers ever need them.

4. Startup cleanup for stale kernel state (nice-to-have)

After container restart, stale ip rules/routes from the previous process remain in the netns. Functionally harmless for traffic correctness:
- nfqlb has no config at startup → packets in nfqueue dropped
- LB readiness gate prevents VIP attraction
- ensureRule reuses existing rules (idempotent with fixed priority)
- RouteReplace overwrites routes when targets are re-learned

Only concern: orphaned state from DGs deleted while LB was down (no reconcile trigger, no finalizers). Minor resource leak, misleading during debugging. A priority-
based sweep at startup would address this but is not required for correctness.

5. Documentation update

docs/operations/constraints-and-limitations.md still documents the old formula (DG_ID × 1024 + 5000) and 1024-endpoint-per-DG hard limit. Needs updating for the new 
dynamic offset allocation.

6. Dual-stack LB controller gap (pre-existing, not a blocker)

newTargets[identifier] = endpoint.Addresses overwrite in targets.go — only last-processed EndpointSlice family survives. Tracked by #144, not in scope for this PR.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━


### Questions for the author

- Is the DeleteTarget stuck-state scenario (item 1) considered acceptable risk, or should there be a mechanism to ensure failed deactivations are retried?
- Was the order-sensitivity of slicesEqual considered? Should it be set-based or sorted?
- Is the deletePolicyRoute early-return on RuleDel failure intentional, or should it always attempt both deletions?
- Was the startup cleanup considered and deferred? If so, a TODO or issue reference would be helpful.

ljkiraly:
1/ Partial failure handling in multi-step DeleteTarget
This issue is not related to the refactoring it was present already. New issue raised #151.

2/ slicesEqual in AddTarget is order-sensitive
Other legacy issue - can be solved by #144


3/ `deletePolicyRoute` is fragile

Solved by commit a3b834d

4/Startup cleanup for stale kernel state (nice-to-have)

Solved by commit cfdb0f8

5/ Documentation update

Solved by commit ff9c17e

6/ Dual-stack LB controller gap (pre-existing, not a blocker)

Will be solved by #144

Issue #108, opened 2026-04-15
LB controller: belongsToGateway does not check DistributionGroup.spec.parentRefs


zolug:
## Problem

The LB controller's `belongsToGateway()` function resolves DistributionGroup-to-Gateway membership **only** via L34Routes. It checks whether any L34Route references both this Gateway (via `parentRefs`) and the DistributionGroup (via `backendRefs`). It does **not** check `DistributionGroup.spec.parentRefs` directly.

This creates an inconsistency: the `gatewayEnqueue` mapper correctly checks `distGroup.Spec.ParentRefs` to trigger reconciliation when the Gateway changes, but the reconcile loop then skips the DG because `belongsToGateway` returns `false`.

**Result:** A DistributionGroup with only a direct `parentRef` to a Gateway (no L34Route linking them) will be enqueued but silently not processed. The NFQLB instance, targets, and flows for that DG are never created.

## Why direct parentRefs matter

The `DistributionGroup.spec.parentRefs` field was designed to support **warm endpoint pooling** (see #6, "The Lazy vs. Eager Reconciliation Challenge"). Without a direct parentRef, the controller cannot resolve the network context or maintain endpoints until an L34Route creates the link. This causes:

- **Convergence delays**: The data plane cannot be pre-programmed before the first L34Route is created
- **Operational churn**: Ephemeral Routes cause the controller to repeatedly teardown and rebuild endpoint slices or hash rings
- **Reverse-lookup overhead**: Frequent backend updates require constant Route-to-Backend resolution

The design intent from #6:
> To address these convergence problems and minimize unnecessary churn, an optional reference toward a Gateway object (e.g., via `parentRef`) might be supported. This allows the controller to determine the network context and maintain a "warm" pool of endpoints independently of the Route lifecycle.

With the current bug, this warm pooling capability is broken at the LB layer — the DistributionGroup controller correctly creates EndpointSlices for DGs with direct parentRefs, but the LB controller ignores them.

## Current behavior

```
gatewayEnqueue:    checks DG.spec.parentRefs → enqueues DG ✓
belongsToGateway:  checks L34Routes only     → returns false, DG skipped ✗
```

## Expected behavior

`belongsToGateway` should return `true` if **either**:
1. An L34Route references both this Gateway and the DG (current check), **or**
2. The DG's `spec.parentRefs` directly references this Gateway

This matches the resolution priority from #6:
1. If `backend.spec.parentRef` is set → Use that Gateway's network
[...]

PR #110, opened 2026-04-16
Fix belongsToGateway to check DistributionGroup.spec.parentRefs

ljkiraly:
belongsToGateway only checked L34Route-based indirect references (L34Route.parentRefs → Gateway AND L34Route.backendRefs → DG). It did not check DistributionGroup.spec.parentRefs directly.

This caused DGs with only a direct parentRef to a Gateway (no L34Route linking them) to be enqueued by the gatewayEnqueue mapper but silently skipped during reconciliation.

Add direct parentRef check as the first path in belongsToGateway, before the L34Route-based check. Uses the same defaulting logic as gatewayEnqueue (default Group to gateway.networking.k8s.io, Kind to Gateway, Namespace to DG namespace).

When a DG has only a direct parentRef (no L34Route), the NFQLB instance and targets are created (warm pool), but reconcileFlows produces zero flows — correct behavior for pre-populated hash ring.

Also extract gateway API string literals into constants.

Related issue: gh-108

### DIFF
--- a/internal/controller/loadbalancer/controller.go
+++ b/internal/controller/loadbalancer/controller.go
@@ -86,7 +86,11 @@ type nftablesManager interface {
 	Cleanup() error
 }
 
-const kindDistributionGroup = "DistributionGroup"
+const (
+	kindDistributionGroup = "DistributionGroup"
+	groupGatewayAPI       = "gateway.networking.k8s.io"
+	kindGateway           = "Gateway"
+)
 
 func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logr := log.FromContext(ctx)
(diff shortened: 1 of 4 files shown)


internal/controller/loadbalancer/controller.go
```diff
@@ -86,7 +86,11 @@ type nftablesManager interface {
 	Cleanup() error
 }
 
-const kindDistributionGroup = "DistributionGroup"
+const (
+	kindDistributionGroup = "DistributionGroup"
+	groupGatewayAPI       = "gateway.networking.k8s.io"
```
zolug: gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
gatewayv1.GroupName

(sigs.k8s.io/gateway-api@v1.4.1/apis/v1/zz_generated.register.go)

ljkiraly: Fixed by 0e0bfce

zolug (approved):

PR #112, opened 2026-04-16
Remove defaultNetlinkOps wrapper in sidecar

### LINKED ISSUE #105: Simplify the netlink interface in sidecar
Should we do a similar simplification like https://github.com/Nordix/Meridio-2/pull/96#discussion_r3064654128 in sidecar too?

I mean, the sidecar package defines a netlinkOps interface for testability, along with a defaultNetlinkOps struct that delegates each method to the corresponding netlink package-level function. This wrapper is unnecessary because *netlink.Handle already satisfies the interface, so the package-level functions themselves could just call methods on &Handle{} internally.

szvincze:
This PR resolves #105. *netlink.Handle already satisfies the netlinkOps interface directly, so the defaultNetlinkOps struct and its 11 forwarding methods are unnecessary. Use &netlink.Handle{} as the production implementation.


bszirtes (approved):
Looks good to me.

zolug (approved):

PR #114, opened 2026-04-20
TCP Authentication Option

elajkat:
- **TCP-AO: Add TCP Authentication Option to GatewayRouter BGP**
- **TCP-AO: add ctraffic to router image**


elajkat (commented):
I used this to test traffic also, so not sure if we really need it or just document somewhere (ie.: docs/router-testing-kind.md ) that hey if you want traffic test, add this to your dockerfile etc...

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -74,6 +74,10 @@ type BgpSpec struct {
 	// BGP listening port of the Attractor FrontEnds.
 	// +optional
 	LocalPort *uint16 `json:"localPort,omitempty"`
+
+	// BGP authentication with TCP Authentication Option (RFC5925).
+	// +optional
+	Authentication *BgpTcpAoSpec `json:"authentication,omitempty"`
```
elajkat, line 81: I am not sure if we have to give option here for the "old/legacy" auth with MD5, or let the only authentication be TCP-AO

hack/vpn-gateway/bird-gw.conf
```diff
@@ -31,8 +31,16 @@ protocol kernel {
 # ns-a gw-a1 peers on VLAN 100 (ASN 64512)
 protocol bgp GW4_A1 from LINK {
 	local 169.254.100.150 port 10179 as 4200000000;
-	neighbor range 0.0.0.0/0 port 10179 as 64512;
-	dynamic name "GW4_A1_";
+	neighbor 169.254.100.1 port 10179 as 64512;
```
elajkat: Yeah, this is the thing I already mentioned during the demo: tcp-ao works only with static routes....
I have to update the series with some documentation for sure where this is also highlighted.

internal/bird/bird_test.go
elajkat, line 1: TODO: add some tests to cover the new functionality

internal/bird/config.go
```diff
@@ -86,7 +88,7 @@ template bgp BGP_TEMPLATE {
 	hold time 90;
 	bfd on;
 	graceful restart off;
-	setkey off;
+	setkey on;
```
elajkat: setkey is off for MD5 but for AO we have to set it

test/examples/sample-router-test.yaml
```diff
@@ -76,7 +76,7 @@ spec:
       privileged: true
   containers:
   - name: router
-    image: registry.nordix.org/cloud-native/meridio-2/router:latest
+    image: 127.0.0.1:5001/router:latest
```
elajkat: uuups, I think this is an accident....

elajkat (commented):
Some sentences I just realised as I revisited....

api/v1alpha1/gatewayrouter_types.go
arp-est: Asked Zoli we don't need the md5 auth part

internal/bird/bird_test.go
elajkat: done

test/e2e/Makefile
bszirtes: Please include the new `tcp-ao` suite here.

test/e2e/suites/tcp-ao/targets.yaml
```diff
@@ -0,0 +1,60 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: target-b
```
bszirtes: Please rename the `target-b` to something else (e.g. `target-t`) as it collides with the common-appnetwork suite's targets, making the tests of that suite fail, when they are simultaneously deployed.

bszirtes:
Please include the `tcp-ao` entry in `.github/workflows/e2e.yaml`, so it will get picked up by the e2e tests:

```yaml
matrix:
 suite:
    - common-appnetwork
    - separate-appnetwork
    - sctp-multihoming
    - low-mtu
    - tcp-ao  # Add this
```

bszirtes:
### Minor Suggestions (can be follow-up PRs)

#### 1. Password escaping for BIRD config

**File**: `internal/bird/config.go:234`

If a password contains quotes or backslashes, it could break BIRD config syntax:
```go
secret "%s";  // Password with " would break this
```

**Suggestion**: Add escaping
```go
func escapeBirdString(s string) string {
   s = strings.ReplaceAll(s, \, \\)
   s = strings.ReplaceAll(s, ", \")
   return s
}
```

#### 2. Missing unit tests for Secret resolution

**File**: `internal/controller/router/controller.go` (resolvePasswords, getTcpAoSecret)

These functions are only tested via E2E happy path. Would be good to add unit tests for:
- Secret not found
- Secret key not found
- Multiple routers with different Secrets

#### 3. Document unused API fields

**File**: `api/v1alpha1/gatewayrouter_types.go:127-134`

`currentKeyId` and `nextKeyId` are defined but not used in config generation yet (all keys marked as `preferred`). Add a comment:
```go
// CurrentKeyId specifies the active key ID for sending.
// TODO: Implement in BIRD config generation (all keys currently marked preferred)
// +optional
CurrentKeyId uint8 json:"currentKeyId,omitempty"
```

test/e2e/Makefile
elajkat: thanks, done

test/e2e/suites/tcp-ao/targets.yaml
elajkat: good catch, done

elajkat:
I added function to escape characters, added a few unit tests and also added the TODO line to the CurrentKeyId and NextKeyId fields

arp-est:
Hi, these are what AI found and seem resonable suggestions:
# TCP-AO Implementation Review

Review of commits `46f146e` (implementation) and `6ba2cf4` (e2e test).

## Dead/Unused Code

### 3. `RNextKeyId` field is never used — OPEN

`TcpAoKeyChain.RNextKeyId` is defined in the API but never referenced in config generation. BIRD's TCP-AO `key {}` block supports `send id` and `recv id` — this field should map to the recv ID in the generated config, or be removed if not needed.

### 4. ~~`CurrentKeyId` and `NextKeyId` are never used~~ ✅ ACKNOWLEDGED

`BgpTcpAoSpec.CurrentKeyId` and `NextKeyId` now have `// TODO` comments indicating they're intentionally deferred. All keys are currently marked `preferred;` as a known limitation.

// Arpi: Also about this, I mean don't we want to support all of the tcpao feature and its options? Why only a TODO comment?

## Logic Issues

### 5. Silent failure in `resolvePasswords` — OPEN

When a secret fetch fails, the code logs the error and `continue`s. This means a misconfigured secret reference results in a BIRD config *without* authentication — the BGP session may still come up unauthenticated (if the peer also falls back), which is the opposite of what the user intended. This should either:
- Fail the reconciliation (return error), or
- Set a degraded status condition on the Gateway

Note: the new controller tests explicitly validate this "skip on failure" behavior (`TestResolvePasswords/SecretNotFound_SkipsKey`, `MultipleKeys_PartialFailure`), so this appears to be an intentional design choice. However it deserves at least a code comment explaining *why* partial degradation is preferred over hard failure.

// Arpi: It would probably be better to fail the reconcile in this case, than to silently disable the tcpao, I mean the users intent was to switch it on, hence its better to fail than silently do the opposite of what was requested :) 

### 6. `setkey` changed from `off` to `on` without clear reason — OPEN

The template changes `setkey off;` → `setkey on;` in the BGP template. `setkey` controls whether BIRD installs TCP-MD5/AO keys into the kernel socket. This is correct for TCP-AO to work, but it's applied *globally* to the template — meaning all BGP sessions now have `setkey on;`, even those without authentication. This shouldn't cause issues in practice (no keys = no effect), but it's a broader change than necessary and should be documented.

// Arpi: If what AI is saying here is true, then I would prefer to leave the old behaviour in place and just give the user the option to set it to on if they want to
## API Design

### 7. `MinItems=1` on an `+optional` field — OPEN

```go
// +kubebuilder:validation:MinItems=1
// +optional
Keychain []TcpAoKeyChain `json:"keychain,omitempty"`
```
This is contradictory: the field is optional (can be omitted), but if present it must have at least 1 item. That's *probably* the intent, but it's worth a comment clarifying it. Also, if `Authentication` is non-nil but `Keychain` is empty/omitted, the validation would reject it — which may confuse users.

### 8. ~~All keys marked `preferred`~~ ✅ ACKNOWLEDGED

See item 4. Known limitation with TODO comments; `CurrentKeyId` logic deferred.

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -103,6 +107,46 @@ type BfdSpec struct {
...
+	// KeyId is the Send_ID for this key (0-255).
+	KeyId uint8 `json:"keyId"`
+
+	// RNextKeyId is the Recv_ID for this key (0-255).
+	RNextKeyId uint8 `json:"rNextKeyId"`
+
+	// Algorithm specifies the cryptographic algorithm.
+	// Valid values: "hmac-sha-1", "hmac-sha-256", "cmac-aes-128"
+	// +kubebuilder:validation:Enum=hmac-sha-1;hmac-sha-256;cmac-aes-128
+	Algorithm string `json:"algorithm"`
```
arp-est, line 141: I've checked bird 3.3.0 docs, and the possible values there seem to be more than 3.
It might be better to just leave it as a simple string, than restricting to a subset, also in the future other crypto algos might emerge
https://bird.nic.cz/doc/bird-3.3.0.html#bgp

> Used cryptographic algorithm can be specified for each key with the option algorithm. Possible values are: hmac md5, hmac sha1, hmac sha224, hmac sha256, hmac sha384, hmac sha512, and cmac aes128. Default value is hmac sha1.

test/e2e/suites/tcp-ao/dg.yaml
```diff
@@ -0,0 +1,26 @@
+---
+apiVersion: meridio-2.nordix.org/v1alpha1
+kind: DistributionGroup
+metadata:
+  name: dg-b1
```
bszirtes: It would be cleaner to rename this to `dg-t1`, because the common-appnetwork suite uses this naming already.

test/e2e/suites/tcp-ao/dg.yaml
```diff
@@ -0,0 +1,26 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: target-tao
+  maglev:
+    maxEndpoints: 32
+  parentRefs:
+  - name: gw-b1
```
bszirtes: Same here.

test/e2e/suites/tcp-ao/gateway.yaml
```diff
@@ -0,0 +1,88 @@
+---
+# Gateway B1 on VLAN 800 with TCP-AO
```
bszirtes: The comment here could also be adjusted when the new `gw` name will be in place.

test/e2e/suites/tcp-ao/gateway.yaml
```diff
@@ -0,0 +1,88 @@
...
+kind: Gateway
+metadata:
+  name: gw-b1
+spec:
+  gatewayClassName: meridio
+  infrastructure:
+    parametersRef:
+      group: meridio-2.nordix.org
+      kind: GatewayConfiguration
+      name: gwconfig-b1
```
bszirtes: Same here.

test/e2e/suites/tcp-ao/routing.yaml
```diff
@@ -0,0 +1,110 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: bgp-tcpao-secret
+type: Opaque
+stringData:
+  master-key-1: "my-secure-master-key-string"
+  master-key-2: "my-new-master-key-for-rotation"
+---
+# gw-b1 router (VLAN 800, ASN 64520)
```
bszirtes: Same here.

test/e2e/suites/tcp-ao/routing.yaml
```diff
@@ -0,0 +1,110 @@
...
+apiVersion: meridio-2.nordix.org/v1alpha1
+kind: GatewayRouter
+metadata:
+  name: gw-b1-router-v4
```
bszirtes: Same here.

test/e2e/suites/tcp-ao/routing.yaml
```diff
@@ -0,0 +1,110 @@
...
+  name: route-b1
+spec:
+  parentRefs:
+  - name: gw-b1
+  backendRefs:
+  - group: meridio-2.nordix.org
+    kind: DistributionGroup
+    name: dg-b1
+  destinationCIDRs:
+  - "30.0.0.1/32"
```
bszirtes: Since these VIPs (`30.0.0.1` and `30.0.0.2`) are already used in the common-appnetwork suite, when running all of the tests simultaneously, this could break the execution. It would better to have a dedicated pair of VIPs for this suite, as for all the others.

zolug:
Looked at it a week ago, felt more like a POC than a production ready PR.

- CRD API and provided functionality is not aligned. E.g. asymmetric per-key directional IDs not supported by the current implementation: BIRD config generation uses only id % (based on key.KeyId), i.e. RNextKeyId can be set in the CRD but is ignored. And key rotation is broken.
- The implementation must not allow in any case falling back to non-authentication mode if configuration explicitly expressed that BGP authentication must be employed for a Gateway Router.
- The router controller cannot handle the use case when a Secret holding an authentication key is added with a delay. It doesn't watch Secrets for example.
- BGP authentication shall not be enabled by default, and thus existing e2e test setups must be obeyed. Consider a separate suite.
- The BIRD BGP `setkey` option is BSD only, isn't it? Why is it required?

elajkat:
> Please include the `tcp-ao` entry in `.github/workflows/e2e.yaml`, so it will get picked up by the e2e tests:
> 
> ```yaml
> matrix:
>  suite:
>     - common-appnetwork
>     - separate-appnetwork
>     - sctp-multihoming
>     - low-mtu
>     - tcp-ao  # Add this
> ```

Done, thanks for the tip

api/v1alpha1/gatewayrouter_types.go
elajkat: Isn't it better to catch it on the API level?

test/e2e/suites/tcp-ao/dg.yaml
elajkat: thanks, good idea, I changed it

test/e2e/suites/tcp-ao/dg.yaml
elajkat: done

test/e2e/suites/tcp-ao/gateway.yaml
elajkat: Done

test/e2e/suites/tcp-ao/gateway.yaml
elajkat: done

test/e2e/suites/tcp-ao/routing.yaml
elajkat: Done

test/e2e/suites/tcp-ao/routing.yaml
elajkat: done

test/e2e/suites/tcp-ao/routing.yaml
elajkat: Thank, good point I changed it

elajkat:
> Looked at it a week ago, felt more like a POC than a production ready PR.
> 
> * CRD API and provided functionality is not aligned. E.g. asymmetric per-key directional IDs not supported by the current implementation: BIRD config generation uses only id % (based on key.KeyId), i.e. RNextKeyId can be set in the CRD but is ignored. And key rotation is broken.
(Lajos): tcpAoConfig now sets preferred also, thanks

> * The implementation must not allow in any case falling back to non-authentication mode if configuration explicitly expressed that BGP authentication must be employed for a Gateway Router.
(Lajos):  With Árpi we added check for resolvePasswords and it will return with error if the parsing is not possible.

> * The router controller cannot handle the use case when a Secret holding an authentication key is added with a delay. It doesn't watch Secrets for example.
(Lajos): watches is added to handle such case

> * BGP authentication shall not be enabled by default, and thus existing e2e test setups must be obeyed. Consider a separate suite.
(Lajos): there is a separate tcp-ao e2e suite now, thanks Botond for checking that as well to be more robust.

> * The BIRD BGP `setkey` option is BSD only, isn't it? Why is it required?
(Lajos): setkey is now added only to per protocol if auth is set for the router

elajkat:
> Hi, these are what AI found and seem resonable suggestions:
> 
> # TCP-AO Implementation Review
> Review of commits `46f146e` (implementation) and `6ba2cf4` (e2e test).
> 
> ## Dead/Unused Code
> ### 3. `RNextKeyId` field is never used — OPEN
> `TcpAoKeyChain.RNextKeyId` is defined in the API but never referenced in config generation. BIRD's TCP-AO `key {}` block supports `send id` and `recv id` — this field should map to the recv ID in the generated config, or be removed if not needed.
(Lajos): thanks, now both recv id and send id set

> 
> ### 4. ~`CurrentKeyId` and `NextKeyId` are never used~ ✅ ACKNOWLEDGED
> `BgpTcpAoSpec.CurrentKeyId` and `NextKeyId` now have `// TODO` comments indicating they're intentionally deferred. All keys are currently marked `preferred;` as a known limitation.
(Lajos): thanks done, now preferred is only added when needed.

> 
> // Arpi: Also about this, I mean don't we want to support all of the tcpao feature and its options? Why only a TODO comment?
> 
> ## Logic Issues
> ### 5. Silent failure in `resolvePasswords` — OPEN
> When a secret fetch fails, the code logs the error and `continue`s. This means a misconfigured secret reference results in a BIRD config _without_ authentication — the BGP session may still come up unauthenticated (if the peer also falls back), which is the opposite of what the user intended. This should either:
> 
> * Fail the reconciliation (return error), or
> * Set a degraded status condition on the Gateway
> 
> Note: the new controller tests explicitly validate this "skip on failure" behavior (`TestResolvePasswords/SecretNotFound_SkipsKey`, `MultipleKeys_PartialFailure`), so this appears to be an intentional design choice. However it deserves at least a code comment explaining _why_ partial degradation is preferred over hard failure.
(Lajos): Done, thanks for checking together.

> 
> // Arpi: It would probably be better to fail the reconcile in this case, than to silently disable the tcpao, I mean the users intent was to switch it on, hence its better to fail than silently do the opposite of what was requested :)
> 
> ### 6. `setkey` changed from `off` to `on` without clear reason — OPEN
> The template changes `setkey off;` → `setkey on;` in the BGP template. `setkey` controls whether BIRD installs TCP-MD5/AO keys into the kernel socket. This is correct for TCP-AO to work, but it's applied _globally_ to the template — meaning all BGP sessions now have `setkey on;`, even those without authentication. This shouldn't cause issues in practice (no keys = no effect), but it's a broader change than necessary and should be documented.
> 
(Lajos): setkey on is now set only for routers for which authentication is set.

> // Arpi: If what AI is saying here is true, then I would prefer to leave the old behaviour in place and just give the user the option to set it to on if they want to
> 
> ## API Design
> ### 7. `MinItems=1` on an `+optional` field — OPEN
> ```go
> // +kubebuilder:validation:MinItems=1
> // +optional
> Keychain []TcpAoKeyChain `json:"keychain,omitempty"`
> ```
> 
> This is contradictory: the field is optional (can be omitted), but if present it must have at least 1 item. That's _probably_ the intent, but it's worth a comment clarifying it. Also, if `Authentication` is non-nil but `Keychain` is empty/omitted, the validation would reject it — which may confuse users.
> 
> ### 8. ~All keys marked `preferred`~ ✅ ACKNOWLEDGED
> See item 4. Known limitation with TODO comments; `CurrentKeyId` logic deferred.

api/v1alpha1/gatewayrouter_types.go
arp-est: Well, it would be nice to catch it on API level, but if bird comes up with a new algorithm every release then that would also be cumbersome to follow up IMO. 
We would have to update the algos in the api every time a bird adds a new one, it might be once every 5 years, or I don't know each release, but I don't want to go through the release notes every time, so at least for now I'm leaning in this case towards the second option.

hack/vpn-gateway/init.sh
```diff
@@ -45,6 +45,16 @@ ip link set vlan7 up
...
+ip addr add 169.254.103.150/24 dev vlan8
+
+# VLAN 900 — tcp-ao gw-b2
+ip link add link eth0 name vlan9 type vlan id 900
+ip link set vlan9 up
+ip addr add 169.254.203.150/24 dev vlan9
+
 ethtool -K eth0 tx off
 
 # VLAN 800 — pod-cache-label gw-pcl
```
arp-est: we messed up the rebase here, Laci reserved vlan 800, we should have probably updated tcpao, to use 900, and 1000

internal/bird/config.go
```diff
@@ -203,9 +206,78 @@ func toRouterData(router *meridio2v1alpha1.GatewayRouter) (routerData, error) {
...
+
+func escapeBirdString(s string) string {
+	s = strings.ReplaceAll(s, `\`, `\\`)
+	s = strings.ReplaceAll(s, `"`, `\"`)
+	s = strings.ReplaceAll(s, "\n", `\n`)
+	s = strings.ReplaceAll(s, "\r", `\r`)
+	return s
+}
+
+func convertAlgorithm(algo string) string {
```
arp-est, line 285: why do we need a convert algo at the first place, can't we just give it to the yaml as it is as a string?

hack/vpn-gateway/init.sh
arp-est: this should be updated too.

internal/bird/bird.go
```diff
@@ -63,9 +63,6 @@ func New(cfg Config) (*Bird, error) {
 	if cfg.SocketPath == "" || cfg.ConfigFile == "" {
 		return nil, fmt.Errorf("bird config: SocketPath and ConfigFile are required")
 	}
-	if cfg.TableID < 1 || cfg.TableID > math.MaxUint32-1 {
```
arp-est, line 66: this probably got removed accidentally during rebase, not that it matters too much but its nice to have.

arp-est: also now that I think about it it would be a bit more robust to make tableId something that can for sure hold number greater than uint32_max, (according google on a 64 bit system int is 64 bit, so this is sort of fine, but not on 32bit systems)
I'm not saying for you to fix it though, its enough to re-include it. just wrote it to note, that I should have been more attentive here.

internal/bird/bird_test.go
```diff
@@ -324,3 +323,207 @@ func uint16Ptr(i uint16) *uint16 {
 func boolPtr(b bool) *bool {
 	return &b
 }
+
+func TestConvertAlgorithm(t *testing.T) {
+	tests := []struct {
+		input string
+		want  string
+	}{
+		{"hmac-sha-1", "hmac sha1"},
```
arp-est, line 333: Just noting it here that the test doesn't include all supported algos, though I'm a bit skeptical whether we need a convert algo in the first place (see the other comment).

internal/bird/config.go
```diff
@@ -203,9 +206,78 @@ func toRouterData(router *meridio2v1alpha1.GatewayRouter) (routerData, error) {
 		BFD:        bfd,
 		HoldTime:   holdTime,
 		IPFamily:   ipFamily,
+		TcpAo:      tcpAo,
 	}, nil
 }
 
+func tcpAoConfig(tcpAo *meridio2v1alpha1.BgpTcpAoSpec, passwords map[uint8]string) string {
```
arp-est, line 246: I think you can do this with go templates similarly as with the bird config itself, that would be more aligned with common practices i think.

arp-est: ai gave me this alternative, I didn't check it for correctness though, just to compare it visally:
```golang
type tcpAoData struct {
	Keys      []tcpAoKeyData
	NextKeyId *uint8
}

type tcpAoKeyData struct {
	SendId    uint8
	RecvId    uint8
	Secret    string
	Algorithm string
	Preferred bool
}

var tcpAoTmpl = template.Must(template.New("tcpao").Funcs(template.FuncMap{
	"deref": func(p *uint8) uint8 { return *p },
}).Parse(`setkey on;
	authentication ao;
	keys {
{{- range .Keys}}
		key {
			send id {{.SendId}};
			recv id {{.RecvId}};
			secret "{{.Secret}}";
			algorithm {{.Algorithm}};
{{- if .Preferred}}
			preferred;
{{- end}}
		};
{{- end}}
	};
{{- if .NextKeyId}}
	rnext id {{deref .NextKeyId}};
{{- end}}`))


func tcpAoConfig(tcpAo *meridio2v1alpha1.BgpTcpAoSpec, passwords map[uint8]string) string {
	if tcpAo == nil || len(tcpAo.Keychain) == 0 {
		return ""
	}

	var keys []tcpAoKeyData
	for _, key := range tcpAo.Keychain {
		password := passwords[key.KeyId]
		if password == "" {
			continue
		}
		keys = append(keys, tcpAoKeyData{
			SendId:    key.KeyId,
			RecvId:    key.RNextKeyId,
			Secret:    escapeBirdString(password),
			Algorithm: convertAlgorithm(key.Algorithm),
			Preferred: tcpAo.CurrentKeyId == nil || *tcpAo.CurrentKeyId == key.KeyId,
		})
	}

	if len(keys) == 0 {
		return ""
	}

	var buf strings.Builder
	if err := tcpAoTmpl.Execute(&buf, tcpAoData{Keys: keys, NextKeyId: tcpAo.NextKeyId}); err != nil {
		return ""
	}
	return buf.String()
}

```

arp-est:
Also here's what the ai found, some of it overlaps with what I found:


>   Review: tcpao3 branch vs. master
> 
>   This branch adds TCP Authentication Option (RFC 5925) support to GatewayRouter BGP sessions — API types, BIRD config generation, controller logic, unit tests, e2e tests, and
>   documentation.
> 
>   ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> 
>   Critical
> 
>   1. VLAN 800 collision in hack/vpn-gateway/init.sh
> 
>   Lines 48–50 create vlan8 (VLAN 800) for tcp-ao, then lines 60–62 (after ethtool) create the same vlan8/VLAN 800 for pod-cache-label. The second ip link add will fail with EEXIST.
>   It works by coincidence since both use the same IP (169.254.103.150/24) and the script doesn't use set -e, but it's architecturally broken — if pod-cache-label ever changes its IP
>   this will silently shadow it.
> 
>   Fix: The pod-cache-label block should be removed (it's now redundant with tcp-ao's block), or tcp-ao should use a distinct VLAN.
> 
>   2. setkey off; removed from the BGP template
> 
>   The template (config.go line 77) no longer emits setkey off; for any sessions. For non-authenticated BGP sessions, the behavior now depends on BIRD's compiled default, which may
>   vary across versions. The TCP-AO path correctly emits setkey on;, but sessions without authentication lose the explicit setkey off;.
> 
>   Fix: Add conditional logic in the template:
> 
>   {{- if .TcpAo}}
>     {{.TcpAo}}
>   {{- else}}
>     setkey off;
>   {{- end}}
> 
>   ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> 
>   Important
> 
>   3. resolvePasswords contradicts documentation
> 
>   The code returns an error when any secret is missing, blocking the entire reconcile for ALL routers (including those without authentication). The docs say: "Keys with missing or
>   unreadable Secrets are skipped (logged as errors) — partial authentication is possible if at least one key resolves."
> 
>   Either the code should skip failed routers and continue (matching the docs), or the docs should state reconciliation is blocked until all secrets resolve.
> 
>   4. RBAC grants list+watch on ALL Secrets
> 
>   config/rbac/lb-serviceaccount.yaml adds secrets to the same rule as pods — get, list, watch on all secrets in the namespace. This is a necessary trade-off for informer-based
>   watches (controller-runtime's .Watches() requires list+watch), but it's worth documenting the security implication in a comment. Consider whether a direct Get + RequeueAfter
>    approach would be acceptable instead.
> 
>   5. secretEnqueue lists ALL GatewayRouters on every Secret event
> 
>   No early filtering. In namespaces with many secrets (common in production), every secret create/update/delete triggers a GatewayRouter list call. Consider adding a predicate
>   filter on the watch to only trigger for secrets that match known names.
> 
>   6. failed_routers uses snake_case
> 
>   Go convention is failedRouters. This will be flagged by linters.
> 
>   7. TableID range validation removed
> 
>   The check if cfg.TableID < 1 || cfg.TableID > math.MaxUint32-1 was removed from bird.go. This appears unrelated to TCP-AO and allows invalid TableID values through. Likely
>   accidental.
> 
>   8. E2E test doesn't validate traffic or authentication enforcement
> 
>   The test verifies BIRD config contains authentication ao; and that the session is Established, but doesn't:
> 
>   - Ping VIPs to confirm traffic flows
>   - Test with a wrong key (negative test)
>   - Prove TCP-AO is actually enforced vs. silently skipped
> 
>   ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> 
>   Minor
> 
>   9. Empty password treated as missing — tcpAoConfig checks if password == "" which treats an empty-string secret value (legitimate, if poor) as missing. Use the comma-ok idiom:
>   password, ok := passwords[key.KeyId]; if !ok { continue }.
> 
>   10. convertAlgorithm doesn't handle all documented algorithms — CRD comment lists cmac-aes-256, umac-64, umac-128 but these have no case in the switch. They'll pass through as-is
>   which won't match BIRD's expected format.
> 
>   11. Error message conflates failure modes — "no secret was found for %d routers" could actually be "key not found in secret" (secret exists, wrong key).
> 
>   12. Misleading comments in init.sh — Lines 48 and 53 say "tcp-ao gw-b1" / "gw-b2" but the actual gateways are gw-t1 / gw-t2.
> 
>   13. Documentation algorithm list mismatch — docs/controllers/router.md lists only 3 algorithms; CRD comment lists 7.
> 
>   14. tcpaotestCases naming — Should be tcpaoTestCases per Go camelCase.
> 
>   15. E2e timeout 4 minutes — Shorter than the standard 5 minutes without explanation. TCP-AO adds crypto handshake overhead so a shorter timeout is counterintuitive.
> 
>   16. Unused secret key — master-key-2 in the Secret is never referenced. Either add a comment or remove it.

zolug:
Please double check and explain why setkey would be required at all.
It's only meant for BSD targets but not Linux. It wasn't required for BGP MD5 support in Meridio v1 either. TCP-AO doesn't require setkey no matter the target OS.

```
6. setkey changed from off to on without clear reason — OPEN
The template changes setkey off; → setkey on; in the BGP template. setkey controls whether BIRD installs TCP-MD5/AO keys into the kernel socket. This is correct for TCP-AO to work, but it's applied globally to the template — meaning all BGP sessions now have setkey on;, even those without authentication. This shouldn't cause issues in practice (no keys = no effect), but it's a broader change than necessary and should be documented.

(Lajos): setkey on is now set only for routers for which authentication is set.
```

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -104,6 +108,48 @@ type BfdSpec struct {
 	Multiplier *uint16 `json:"multiplier,omitempty"`
 }
 
+// BgpTcpAo defines the parameters to configure TCP Authentication Option (RFC5925).
+type BgpTcpAoSpec struct {
+	// KeyChain defines the list of TCP-AO keys for authentication and rotation.
+	// When Authentication is specified, at least one key must be provided.
+	// +kubebuilder:validation:MinItems=1
+	// +optional
```
zolug: With `optional` marker this translates to that even if Keychain field is absent from the configured BgpTcpAoSpec it still passes. Yet, at least one key must be provided if Authentication is specified. The parent BgpTcpAoSpec is already optional, if authentication is not specified, the BgpTcpAoSpec should be omitted from the config. But if the block is included, keys are mandatory, aren't they?

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -104,6 +108,48 @@ type BfdSpec struct {
...
+
+	// NextKeyId specifies the key ID the peer should transition to (maps to BIRD's rnext id).
+	// +optional
+	NextKeyId *uint8 `json:"nextKeyId,omitempty"`
+}
+
+// TcpAoKeyChain defines a single TCP-AO key configuration.
+type TcpAoKeyChain struct {
+	// KeyId is the Send_ID for this key (0-255).
+	KeyId uint8 `json:"keyId"`
```
zolug: KeyId and RNextKeyId are really confusing (especially if we consider the BgpTcpAoSpec level CurrentKeyId and NextKeyId fields as well).
Why not call these per-chain fields what they are (e.g. SendId and RecvId)? AFAIK the functionality these two are to provide is not BIRD specific, or is it?

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -104,6 +108,48 @@ type BfdSpec struct {
...
+// BgpTcpAo defines the parameters to configure TCP Authentication Option (RFC5925).
+type BgpTcpAoSpec struct {
+	// KeyChain defines the list of TCP-AO keys for authentication and rotation.
+	// When Authentication is specified, at least one key must be provided.
+	// +kubebuilder:validation:MinItems=1
+	// +optional
+	Keychain []TcpAoKeyChain `json:"keychain,omitempty"`
+
+	// CurrentKeyId specifies the active key ID for sending (only this key gets preferred; in BIRD).
+	// If unset, all keys are marked preferred.
```
zolug: What's the point of marking all keys preferred if unset versus not marking them? Wouldn't bird treat a multi-key setup the same in both case?

internal/controller/router/controller.go
```diff
@@ -87,6 +88,11 @@ func (r *RouterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, fmt.Errorf("failed to get gateway routers: %w", err)
 	}
 
+	passwords, err := r.resolvePasswords(ctx, gatewayRouters)
```
zolug: I think this current "stalled reconciliation" approach upon partial password config is acceptable for now. But, don't return an error here, and don't request a re-queue either. Simply retain old config. But feel free to emit an info log.

Rationale:
- partial keychain resolution creates ambiguity about user intent (e.g., currentKeyId references a missing key — what's the "correct" partial config?). Safer to retain the old working config until all dependencies are satisfied.
- since all dependencies are watched (Secrets, GatewayRouters), returning error adds unnecessary backoff delay. The watch events already trigger re-reconcile promptly when the missing Secret appears (or GwRouter config is updated).

Document: The behaviour and to "Update Secrets before modifying GatewayRouter authentication config. Allow informer cache propagation time between changes."

Important: the router controller must NOT update GatewayRouter status to indicate auth issues. Multiple LB Pods each run independent router controllers watching the same GatewayRouter. If one Pod has cache lag (Secret not yet visible) while another resolves fine, multiple writers would fight over the status — causing flapping, 409 Conflicts, and meaningless conditions.

Future improvements for reconciliation and partial config (not required for this PR):
- **CM-level validation for observability:** The Controller Manager (single instance) could validate auth dependencies and set GatewayRouter status conditions (e.g., AuthenticationDependenciesReady=False). This gives operators visibility without the router controller making policy decisions or writing status. While router controllers remain consumers only.

- **Stateful per-router fallback (if blocking proves undesirable):** If user feedback shows one faulty GatewayRouter blocking all others is a real operational problem, the router controller could cache last-working GatewayRouter configs per-router in memory. On faulty config: use cached version for that router, proceed normally with others. In-memory is sufficient since router controller and BIRD share a lifecycle (Pod restart = fresh start, first reconcile rebuilds from scratch).

hack/vpn-gateway/bird-gw.conf
```diff
@@ -31,8 +31,16 @@ protocol kernel {
 # separate-appnetwork gw-a1 peers on VLAN 100 (ASN 64512)
 protocol bgp GW4_A1 from LINK {
 	local 169.254.100.150 port 10179 as 4200000000;
-	neighbor range 0.0.0.0/0 port 10179 as 64512;
-	dynamic name "GW4_A1_";
+	neighbor 169.254.100.1 port 10179 as 64512;
```
zolug: I guess this should be reverted to not break legacy suites. (tcp-ao would need to use its dedicated session config in a separate subnet)

docs/controllers/router.md
```diff
@@ -109,6 +109,63 @@ Blackhole routes (`0.0.0.0/0` and `::/0`) in table 4097 prevent traffic leaking
...
+```
+
+**Supported algorithms**:
+- `hmac-sha-1` — HMAC-SHA1
+- `hmac-sha-256` — HMAC-SHA256
+- `cmac-aes-128` — CMAC-AES128
+
+**Key rotation**: Multiple keys can be configured in the keychain. The `currentKeyId` and `nextKeyId` fields control which key is active for sending and which is advertised for rotation. Both peers must share the same key material for the corresponding key IDs.
+
+**BIRD config output**: The controller generates BIRD `authentication ao` blocks with key entries for each keychain item whose Secret is successfully resolved. Keys with missing or unreadable Secrets are skipped (logged as errors) — partial authentication is possible if at least one key resolves.
```
zolug: Is the following statement correct though? The reconciliation rejects all changes if any dependency was missing doesn't it? The bird pkg would skip partial keys though, but not the reconciler layer above.

`Keys with missing or unreadable Secrets are skipped (logged as errors) — partial authentication is possible if at least one key resolves.`

api/v1alpha1/gatewayrouter_types.go
elajkat: True, I removed optional

api/v1alpha1/gatewayrouter_types.go
elajkat: True, I changed them to sendid and recvid

api/v1alpha1/gatewayrouter_types.go
elajkat: Bird selects the 1st key if none is selected, to avoid relying on BIRD's logic I set preferred if currentkeyid is not set

elajkat:
I will come back to continue with the fixes on Monday(I hope)

hack/vpn-gateway/init.sh
elajkat: thanks, good catch, done

internal/bird/config.go
elajkat: It is not functional requirement, but rfcs use similar naming (ie.: https://www.rfc-editor.org/info/rfc4868/ ), so seemed logical to have it here also

docs/controllers/router.md
elajkat: thanks, I updated it

hack/vpn-gateway/bird-gw.conf
elajkat: done

internal/controller/router/controller.go
elajkat: For now just log if resolvePasswords fails

test/e2e/e2e_suite_tcpao_test.go
```diff
@@ -0,0 +1,199 @@
...
+		targetApp:      "target-tao",
+		targetReplicas: 2,
+		gateways: []gwTestCase{
+			{name: "gw-t1", vip: "60.0.0.1", targets: 2, dgName: "dg-t1"},
+			{name: "gw-t2", vip: "60.0.0.2", targets: 2, dgName: "dg-t2"},
+		},
+	},
+}
+
+var _ = Describe("E2E TCP-AO Test Suite", func() {
```
elajkat: hmmm, I left out the Label("ipv4") param

api/v1alpha1/gatewayrouter_types.go
zolug: But what's the difference compared to not marking all keys as preferred if currentKeyId is not set?

elajkat: It is True, for  the current Bird version it makes no difference, however I prefer setting it explicitly if we bump Bird version and they change this behaviour make it seemless from end user perspective.
Marking all keys "preferre"d when "currentKeyId" is unset is effectively the same as marking none. BIRD selects the first key anyway.

If we go to don't set preferred if CurrentKeyId isn't set I suppose the easiest is in internal/bird/config.go:
- Preferred: tcpAo.CurrentKeyId == nil || *tcpAo.CurrentKeyId == key.SendId,
+ Preferred: tcpAo.CurrentKeyId != nil && *tcpAo.CurrentKeyId == key.SendId,

rubasov:
I scanned this again for unanswered comments and responded where I could, but I'm far from sure I found all. This PR conversation is very noisy. Let me know if you had a concern I did not address in #212 .

elajkat:
https://github.com/Nordix/Meridio-2/pull/212 (tcp ao continued) took over most of this PR

Issue #116, opened 2026-04-28
Externalize hardcoded router controller configuration


arp-est:
## Externalize hardcoded router controller configuration

Several values in the router controller are currently hardcoded as constants. They should be exposed as CLI flags with environment variable binding (following the existing MERIDIO_* pattern and flag precedence:
flags > env > defaults).

### 1. Kernel routing table ID

Current: defaultKernelTableID = 4096 in internal/bird/config.go, kernelTableID = 4096 and blackholeKernelTableID = 4097 in internal/bird/routing.go

Proposed: Single --table-id flag / MERIDIO_TABLE_ID env var (default 4096). Blackhole table is implicitly tableID + 1 (same pattern as Meridio v1's TABLE_ID).

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

### 2. Policy routing rule priorities

Current: rulePriority = 100 and blackholeRulePriority = 101 in internal/bird/routing.go

Proposed: --rule-priority / MERIDIO_RULE_PRIORITY (default 100). Blackhole priority is implicitly priority + 1.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

### 3. BIRD paths

Current: Hardcoded in bird.go New(): socket /var/run/bird/bird.ctl, config /etc/bird/bird.conf

Proposed:
- --bird-socket / MERIDIO_BIRD_SOCKET (default /var/run/bird/bird.ctl)
- --bird-config / MERIDIO_BIRD_CONFIG (default /etc/bird/bird.conf)

Meridio v1 had these as BIRD_COMMUNICATION_SOCK and BIRD_CONFIG_PATH.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

### 4. BGP monitoring interval

Current: 1 * time.Second hardcoded in monitorConnectivity() in cmd/router/cmd/run.go

Proposed: --monitor-interval / MERIDIO_MONITOR_INTERVAL (default 1s).

PR #117, opened 2026-04-29
Add scan time to BIRD kernel protocol configuration

### LINKED ISSUE #80: Add scan time to BIRD kernel protocol configuration
In Meridio v1, BGP routes learned from peers appeared in the kernel routing table with a 30-50 second delay when using BIRD 3.x. The fix was adding `scan time 10` to the kernel protocol blocks that export BGP-learned routes to the kernel: https://github.com/Nordix/Meridio/pull/607

The current Meridio-2 bird.conf template does not set scan time on the kernel protocol blocks, using BIRD's default (60 seconds). Apply the same fix. Make it configurable (--bird-kernel-scan-time / MERIDIO_BIRD_KERNEL_SCAN_TIME, default 10) as Meridio v1 did.

Applies to both IPv4 and IPv6 kernel protocol blocks in internal/bird/config.go.

ljkiraly:
Add configurable scan time to both IPv4 and IPv6 kernel protocol blocks in the BIRD configuration. This fixes a 30-50 second delay for BGP-learned routes to appear in the kernel routing table when using BIRD 3.x.

Configuration:
  --bird-kernel-scan-time / MERIDIO_BIRD_KERNEL_SCAN_TIME (default: 10)

Port of Meridio v1 fix: https://github.com/Nordix/Meridio/pull/607

Closes #80

### DIFF
--- a/internal/bird/config.go
+++ b/internal/bird/config.go
@@ -27,10 +27,9 @@ import (
 )
 
 const (
-	defaultKernelTableID  = 4096
-	defaultKernelScanTime = 10
-	defaultLocalPort      = 179
-	defaultRemotePort     = 179
+	defaultKernelTableID = 4096
+	defaultLocalPort     = 179
+	defaultRemotePort    = 179
 )
 
 type birdConfigData struct {
--- a/internal/bird/config.go
+++ b/internal/bird/config.go
@@ -27,17 +27,19 @@ import (
 )
 
 const (
-	defaultKernelTableID = 4096
-	defaultLocalPort     = 179
-	defaultRemotePort    = 179
+	defaultKernelTableID  = 4096
+	defaultKernelScanTime = 10
+	defaultLocalPort      = 179
+	defaultRemotePort     = 179
 )
 
 type birdConfigData struct {
-	KernelTableID int
-	LogParams     BirdLogParams
-	IPv4VIPs      []string
-	IPv6VIPs      []string
-	Routers       []routerData
+	KernelTableID  int
+	KernelScanTime int
+	LogParams      BirdLogParams
+	IPv4VIPs       []string
+	IPv6VIPs       []string
+	Routers        []routerData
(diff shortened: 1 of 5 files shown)


arp-est (approved):

internal/bird/config.go
```diff
@@ -27,17 +27,19 @@ import (
 )
 
 const (
-	defaultKernelTableID = 4096
-	defaultLocalPort     = 179
-	defaultRemotePort    = 179
+	defaultKernelTableID  = 4096
+	defaultKernelScanTime = 10
```
zolug: Don't use hard-coded default values.
There's a pending issue targeting the removal of these: https://github.com/Nordix/Meridio-2/issues/116

ljkiraly: The bird package should have a sensible default regardless of whether the caller passes the option. This is the standard Go pattern — the package owns its defaults, the caller will override:
`birdInstance := bird.New(bird.WithLogParams(cfg.BirdLogs), bird.WithKernelScanTime(cfg.BirdKernelScanTime))`

The default is eventually defined in internal/common/config/router.go line 66:
fs.IntVar(&c.BirdKernelScanTime, "bird-kernel-scan-time", 10, ...

zolug: > The bird package should have a sensible default regardless of whether the caller passes the option. This is the standard Go pattern — the package owns its defaults

IMHO applies when a package is used by multiple callers or externally. But internal/bird is only used by the router, there's no external consumer. The "sensible default" in the package is redundant when the only caller always passes the value via WithKernelScanTime(cfg.BirdKernelScanTime).

The cobra config layer is the single source of truth for all runtime parameters. Why risk that the two defaults could diverge overtime? What would be the benefit of having defaults in bird package as well?

ljkiraly: Solved by 81f3b85

Issue #122, opened 2026-04-30
Readiness gates: Router controller — connectivity status exposure


ljkiraly:
Step 2 of #90.

The router controller sets Pod readiness gate conditions based on per-IP-family BGP connectivity status. Gates are set to False on startup, True after damped up-transition, and immediately False on session loss.

See #90 "2. Router Controller — Connectivity status exposure" for full description.

**Depends on:** nothing (can be developed in parallel with #121)
**Blocks:** #90 step 3

Issue #121, opened 2026-04-30
Readiness gates: Gateway controller — gate declaration


ljkiraly:
Step 1 of #90.

The Gateway controller declares per-IP-family readiness gates on the LB Deployment pod template based on GatewayConfiguration.spec.networkSubnets CIDRs, and injects POD_NAME/POD_NAMESPACE/POD_UID Downward API env vars into the router container.

See #90 "1. Gateway Controller — Readiness gate declaration" for full description.

**Depends on:** nothing
**Blocks:** #90 step 3

Issue #90, opened 2026-04-01
External Connectivity Exposure via Readiness Gates


zolug:
### Problem

The controller-manager's ENC controller builds next-hop lists from all Running LB Pods (getSLLBRNextHops), without considering whether each LB Pod's router has established BGP sessions with external peers. If an LB Pod loses external connectivity (BGP sessions go down), application Pods continue routing return traffic through it, causing traffic black-holing.

See: constraints-and-limitations.md "No connectivity-based readiness signaling to the controller-manager"

### Goal

Expose per-LB-Pod, per-IP-family external connectivity status via Pod readiness gates. The ENC controller uses this — combined with container-level readiness — to exclude LB Pods without external connectivity from next-hop lists, ensuring application Pods only route return traffic through LB Pods that can actually reach external peers.

#### Why Readiness Gates (not CR status)

GatewayRouter is a shared cluster resource — N LB Pods reconcile the same GatewayRouter CRs. Per-Pod connectivity state belongs on the Pod, not the CR. Readiness gates are the Kubernetes-native mechanism for external controllers to signal per-Pod readiness aspects.

This pattern is established in the ecosystem:
- **Istio** uses istio.io/dataplane-ready to signal sidecar proxy readiness independently from the app container
- **AWS ALB Ingress Controller** uses target-health.alb.ingress.k8s.aws per target group to signal backend health independently
- **KEP-3085 (Pod readiness gates)** was designed for exactly this: external controllers signaling partial readiness

#### Why per-IP-family gates (not a single gate)

IPv4 and IPv6 BGP sessions are independent — one may be up while the other is down. A single gate would force an all-or-nothing decision: either advertise the LB Pod as a next-hop for both families or neither. Per-IP-family gates allow the ENC controller to include an LB Pod in IPv4 next-hops while excluding it from IPv6 (or vice versa).

Note: In Meridio v1, BGP sessions were also independent per IP family, but the data plane was tightly coupled through Network Service Mesh, making it impossible to selectively cease accepting traffic for a single IP version. In Meridio-2, the data plane (secondary network connectivity) is not managed by the operator — it relies on standard L2 adjacency via Multus/NAD — so per-IP-family control is achievable at the routing level.

#### Why not Pod-level Ready as a filter

[...]

PR #125, opened 2026-05-04
Gateway controller: declare per-IP-family readiness gates on LB Deploy...

ljkiraly:
Implement step 1 of gh-90 (External Connectivity Exposure via Readiness Gates). The Gateway controller now:

- Determines required IP families from GatewayConfiguration.spec.networkSubnets
- Declares readiness gates on the LB Deployment pod template:
  - meridio-2.nordix.org/ipv4-connectivity (if IPv4 subnets configured)
  - meridio-2.nordix.org/ipv6-connectivity (if IPv6 subnets configured)
- Injects POD_NAME, POD_NAMESPACE, POD_UID Downward API env vars into the router container (needed for router to patch its own Pod status)

The gates are declared by the Gateway controller but will be set by the router controller (step 2, gh-122).

Ref: gh-121

### DIFF
--- a/internal/controller/gateway/deployment.go
+++ b/internal/controller/gateway/deployment.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"maps"
+	"net"
 	"strings"
 
 	netdefv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
@@ -175,6 +176,8 @@ func reconcileDeploymentSpec(base, template *appsv1.Deployment, gw *gatewayv1.Ga
 		}
 		applyGatewayConfiguration(desired, gwConfig, templateNADAnnotation, base == nil)
 	}
+	// Inject downward API env vars for router Pod identity (POD_NAME, POD_NAMESPACE, POD_UID) - after applying GatewayConfiguration in case it changes the container spec
+	injectRouterPodIdentityEnvVars(desired)
 
 	return desired
 }
@@ -280,6 +283,9 @@ func applyGatewayConfiguration(deployment *appsv1.Deployment, gwConfig *meridio2
 
 	// Network attachments
 	applyNetworkAttachments(deployment, templateNADAnnotation, gwConfig.Spec.NetworkAttachments)
+
+	// Readiness gates (per-IP-family based on InternalSubnets)
+	applyReadinessGates(deployment, gwConfig)
 }
 
 // applyVerticalScaling applies container resources from VerticalScaling config.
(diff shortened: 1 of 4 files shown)


internal/controller/gateway/deployment.go
```diff
@@ -480,3 +483,102 @@ func parseNetworkAnnotation(annotation string) []*netdefv1.NetworkSelectionEleme
 
 	return elements
 }
+
+// ipFamiliesFromInternalSubnets scans CIDRs in gwConfig.Spec.InternalSubnets, returns (hasIPv4, hasIPv6 bool).
+// Each InternalSubnet entry contains a single CIDR of one IP family. Dual-stack uses two separate entries.
+func ipFamiliesFromInternalSubnets(subnets []meridio2v1alpha1.InternalSubnet) (bool, bool) {
```
arp-est, line 492: I think we have a similar function in the router [here](https://github.com/Nordix/Meridio-2/blob/388dd67fc00d3fe8190e5d8ec678eb9505b156c7/internal/bird/config.go#L219), though that only checks whether a single ip or cidr is ipv4 or 6, regardless it might be a bit more sophisticated, to parse the cidr, then just grep for ':'.

ljkiraly: It's not so sophisticated if you don't need the actual CIDR/IP and don't validate anything :)  (no warning log or failure handling in case of invalid addresses either, so I don't think it's better).

arp-est (approved):

internal/controller/gateway/deployment.go
```diff
@@ -174,7 +174,10 @@ func reconcileDeploymentSpec(base, template *appsv1.Deployment, gw *gatewayv1.Ga
 			templateNADAnnotation = template.Spec.Template.Annotations[netdefv1.NetworkAttachmentAnnot]
 		}
 		applyGatewayConfiguration(desired, gwConfig, templateNADAnnotation, base == nil)
+		applyReadinessGates(desired, gwConfig)
```
zolug: Logically, readiness gates are part of "applying the gateway configuration to the deployment".
Any reason why it was not added as a sub-function to applyGatewayConfiguration?

Moving it inside applyGatewayConfiguration (as a call at the end) would:
- Make the dependency on gwConfig implicit
- Group all gwConfig-driven Deployment mutations in one place
- Reduce the surface of reconcileDeploymentSpec

The separation of the function is fine, it's the call placement that seems awkward.

zolug: Otherwise, the PR looks good to me.

internal/controller/gateway/deployment.go
```diff
@@ -480,3 +483,102 @@ func parseNetworkAnnotation(annotation string) []*netdefv1.NetworkSelectionEleme
...
+		}
+	}
+
+	return hasIPv4, hasIPv6
+}
+
+// applyReadinessGates sets deployment.spec.Template.Spec.ReadinessGates authoritatively based on presence of IP families in GatewayConfiguration.
+// Always replaces (controller owns these gates).
+func applyReadinessGates(deployment *appsv1.Deployment, gatewayConfig *meridio2v1alpha1.GatewayConfiguration) {
+	hasIPv4, hasIPv6 := ipFamiliesFromInternalSubnets(gatewayConfig.Spec.InternalSubnets)
```
zolug, line 514: Deriving IP families from internalSubnets is acceptable for now, but has a known trade-off: temporarily removing a subnet entry e.g. to fix a CIDR typo would remove the corresponding readiness gate. Which would trigger a Deployment rollout leading to disruption, even though the IP family intent didn't change.

A dedicated field (decoupling family intent from subnet config) would avoid this churn, but that decision is anyways deferred to #128 (IP family mode management study). The refactor would be minimal if needed later.

ljkiraly (commented):
It is planned to consolidate readiness gate constants into a shared package: `internal/common/constants/readiness-gates.go`

zolug (approved):

Issue #120, opened 2026-04-30
Enforce single IP family per NetworkSubnet in GatewayConfiguration


ljkiraly:
### Problem

The `GatewayConfiguration.spec.networkSubnets` field allows mixing IPv4 and IPv6 CIDRs within a single `NetworkSubnet` entry. The intended semantic is that each `NetworkSubnet` represents a single IP family — dual-stack deployments use two separate entries (one IPv4, one IPv6). This is not enforced at the API level, leading to potential misconfiguration and ambiguity.

The readiness gate feature (issue #90) relies on detecting IP families from `networkSubnets` to declare per-IP-family readiness gates. Enforcing single-family entries makes this detection unambiguous and aligns the API with its intended usage.

### Goal

1. Add CEL validation ensuring all CIDRs within a single `NetworkSubnet` share the same IP family
2. Add CEL validation preventing duplicate IP families across `networkSubnets` entries (e.g., two IPv4 entries)
3. Update documentation and samples to clarify the one-family-per-entry semantic

### Proposed CEL Rules

**On `NetworkSubnet` (single family within entry):**
```
+kubebuilder:validation:XValidation:rule="self.cidrs.all(c, cidr(c).ip().family() == cidr(self.cidrs[0]).ip().family())",message="All CIDRs within a networkSubnet must be the same IP family (use separate networkSubnets for dual-stack)"
```

**On `GatewayConfigurationSpec.networkSubnets` (no duplicate families across entries):**
```
+kubebuilder:validation:XValidation:rule="self.size() <= 1 || cidr(self[0].cidrs[0]).ip().family() != cidr(self[1].cidrs[0]).ip().family()",message="Each networkSubnet must represent a different IP family"
```

### Documentation Changes

- Clarify in `docs/controllers/gateway.md` that each `NetworkSubnet` entry represents one IP family
- Update the GatewayConfiguration field comment in `api/v1alpha1/gatewayconfiguration_types.go`
- Ensure all samples (`config/samples/`, manual testing sections in docs) follow the pattern:

```yaml
networkSubnets:
  - attachmentType: NAD
    cidrs: ["192.168.100.0/24"]       # IPv4 entry
  - attachmentType: NAD
    cidrs: ["2001:db8:100::/64"]      # IPv6 entry (separate)
```

### Breaking Change Consideration

[...]

PR #126, opened 2026-05-05
Enforce single IP family per NetworkSubnet in GatewayConfiguration

ljkiraly:
Add CEL validation rules to ensure:
- All CIDRs within a NetworkSubnet are the same IP family
- No two NetworkSubnet entries share the same IP family

Also adds MaxLength=43 on CIDR strings (longest valid CIDR) to satisfy CEL cost estimator, and updates field documentation to clarify the one-family-per-entry semantic.

Ref: gh-120


api/v1alpha1/gatewayconfiguration_types.go
```diff
@@ -32,13 +32,15 @@ type GatewayConfigurationSpec struct {
 	// +kubebuilder:validation:MaxItems=10
 	NetworkAttachments []NetworkAttachment `json:"networkAttachments"`
 
-	// networkSubnets identifies the subnet(s) where application endpoint IPs reside.
+	// internalSubnets identifies the subnet(s) where application endpoint IPs reside.
 	// Used by the ENC controller to match secondary interfaces in application Pods
```
zolug, line 36: Although the base text did not mention it either, it's also used by the custom endpoint management (DG controller). Just highlighted, because 'ADR-002: Rename NetworkSubnet to InternalSubnet' did not list it either.

zolug (approved):

Issue #87, opened 2026-03-31
Router VIP Exposure Gated on LB Distribution Readiness


zolug:
### Problem

The router container announces known VIPs via BGP as soon as BGP sessions are established, and VIPs learned later are also only tied to BGP session status — without checking whether the collocated LB container has active endpoints to distribute traffic to.
This might cause traffic loss during LB Pod startup or scale-out: external traffic is attracted before the LB can forward it.

See: constraints-and-limitations.md

### Goal

The router controller must gate VIP advertisement on the collocated LB container's distribution readiness. VIPs should only be exposed via BIRD when the LB has at least one DistributionGroup with ready endpoints.

### Mechanism

The LB controller already writes per-DistributionGroup readiness files to a shared emptyDir volume:
- Creates: /var/run/meridio/lb-ready-<distgroup-name> when a DG has ready endpoints
- Removes: when a DG is deleted or has no ready endpoints
- Cleans up all files on startup

The router container currently does not mount this volume.

### Implementation Steps

1. Mount the shared volume in the router container
   - Add lb-run volumeMount at /var/run/meridio (readOnly) to the router container in config/templates/lb-deployment.yaml

2. Watch readiness files as an event source
   - Use controller-runtime's WatchesRawSource builder option to integrate file system events (e.g., via fsnotify) as a non-CR event source into the existing reconcile loop
   - File creation/deletion triggers a reconcile of the Gateway

3. Gate VIP exposure on readiness
   - During reconcile, check for presence of lb-ready-* files in /var/run/meridio/
   - If no readiness files exist: configure BIRD with empty VIP list (no routes advertised)
   - If at least one readiness file exists: configure BIRD with VIPs from Gateway.status.addresses as today

4. Incremental approach
   - Start with a simple file presence check (no fsnotify): poll or check during reconcile
   - Evolve to WatchesRawSource + fsnotify for event-driven reactivity
   - For initial development, a dummy file can replace actual LB input to unblock router-side work

### Considerations

- The readiness directory path should be configurable (flag/env), matching the LB controller's readinessDir
- The router should handle the case where the shared volume is not mounted (backward compatibility / standalone testing) — default to "LB always ready" behavior
- Per-DG granularity is available (individual lb-ready-<dg> files) but for first release, any readiness file present = ready is sufficient
[...]

PR #130, opened 2026-05-08
Expose VIPs only in router if lb status is ready

arp-est:
https://github.com/Nordix/Meridio-2/issues/87


internal/controller/router/controller.go
```diff
@@ -93,11 +105,21 @@ func (r *RouterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		Watches(&meridio2v1alpha1.GatewayRouter{}, handler.EnqueueRequestsFromMapFunc(r.gatewayRouterEnqueue))
+
+	if r.LBReadinessPath != "" {
+		ch := make(chan event.GenericEvent, 1)
+		_ = mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+			return r.watchLBReadinessDir(ctx, ch)
+		}))
+		builder = builder.WatchesRawSource(source.Channel(ch, &handler.EnqueueRequestForObject{}))
+	} else {
+		logf.Log.WithName("setup").Info("WARNING: --lb-readiness-path/MERIDIO_LB_READINESS_PATH not set, VIPs will be advertised without waiting for LB targets")
```
ljkiraly: (WARNING message at Info level?)
The log on empty path is misleading. The default value is /var/run/meridio (non-empty), so the warning only fires if someone explicitly sets --lb-readiness-path="". The log message says "not set" but it's actually "explicitly cleared." The message should say "disabled" rather than "not set."

ljkiraly (changes requested):
### Recommendations

1. Update `docs/controllers/router.md` — move LB Readiness Gating from "Future Work" to an implemented section with configuration details.

2. Add integration test in `controller_test.go` that exercises the full reconcile path with readiness gating 'enabled/disabled'.

#### Coverage Gaps

2.1. No test for the Reconcile integration. The existing `controller_test.go` doesn't test the readiness gating path. Need a test that:

- Sets `LBReadinessPath` to a temp dir
- Verifies VIPs are suppressed when dir is empty
- Verifies VIPs are passed through when readiness file exists

2.2. No test for `LBReadinessPath = "" (disabled mode)`. Should verify VIPs are always advertised regardless of directory state.

2.3. No test for `SetupWithManager` branching. The conditional watcher setup based on `LBReadinessPath` is untested.

2.4. No test for the default: drop in channel send. Should verify the system self-heals when events are dropped.

3. Consider making LB readiness dir configurable on the LB side too (via `--readiness-dir` flag), for consistency and testability. Currently it's a package-level var that tests override directly.
   Asymmetry in path configuration: The LB controller uses a hardcoded `var readinessDir = "/var/run/meridio"` (package-level variable, not configurable via flags), while the Router controller introduces a configurable `--lb-readiness-path` flag. If someone changes the router's path without changing the LB's hardcoded path, the system silently breaks.

4. Fix the WARNING log message — change "not set" to "disabled" since the default is non-empty.

ljkiraly:
A new follow-up issue should be raised for hard-code the readiness directory path. This is an internal communication channel that should not be set by the user. 

-  If we want to make possible for the user to be able to configure the ephemeral volume that should be done by 'GatewayConfiguration' (similar to auto-scaling).
- If there is a need to enable/disable this feature (for e.g. in case of Static-BFD routing) use a separate flag/parameter (and configuration parameter in CR) for this purpose. Currently the empty readiness directory path disable it.

---

Future improvement: Add per-IP-family granularity. Meridio v1 tracks IPv4 and IPv6 connectivity separately. Currently we use a single boolean/flag (any lb-ready-* file exists). If the LB has IPv4 targets but not IPv6, all VIPs (including IPv6) are still advertised. This is acceptable for now, but should be noted.

ljkiraly (approved):
Implementation is clean now.

docs/controllers/router.md
```diff
@@ -155,14 +166,6 @@ Meridio v1 exposes per-GatewayRouter metrics and monitors BIRD route counts:
 - **Memory monitoring**: `birdc show memory` for BIRD memory usage tracking.
 - These should be exposed as Prometheus metrics via the controller-runtime metrics server (already wired but unused).
 
-### LB Readiness Gating (post-MVP)
```
zolug, line 158: should be marked resolved in the limitations doc as well:

```
**10. VIPs advertised regardless of LB distribution readiness**

VIPs are announced via BGP as soon as the router's BGP sessions are established, without waiting for the collocated LB to have active endpoints. This can cause traffic loss during LB Pod startup or scale-out: external traffic is attracted before the LB is capable of distributing it to application endpoints.
```

Otherwise looks good to me.

zolug (approved):

PR #132, opened 2026-05-11
Fix LB routing: use RouteReplace to handle stale routes

### LINKED ISSUE #129: LB controller fails to update fwmark routes when target IP changes
## Description

When a Maglev fwmark is reassigned to a different target IP (e.g., Pod replacement, secondary IP change), the LB controller's AddRoute fails with EEXIST and the fwmark ip rule is never created.

This can also occur if the LB controller misses intermediate reconciliation steps — for example, when application targets scale-in and then out in quick succession. If the controller only reconciles the end result, it sees a fwmark now pointing to a different IP without having processed the intermediate removal that would have cleaned up the old route. The stale route blocks the new one.

Similarly, if a DistributionGroup is deleted and re-created, the fresh Maglev allocation has no existing assignments to preserve instead orders Pods by creation timestamp. If Pods were replaced between the original allocation and the re-creation, their position in the sort order differs, potentially assigning different IPs to the same fwmarks.

## Root Cause

Two related problems in internal/controller/loadbalancer/routing.go:

### 1. deleteRouteInternal uses the wrong IP for cleanup

AddRoute(fwmark, newIP) calls deleteRouteInternal(fwmark, newIP) to clean up before adding. This constructs a netlink.Route{Gw: newIP, Table: fwmark} and calls RouteDel. The kernel only deletes a route if the gateway matches — but the existing route has the old IP as gateway, so the delete is a no-op. The subsequent RouteAdd fails because the table already has a default route (with the old gateway). On failure, the rule added just before is rolled back, leaving no fwmark rule.

### 2. isValidRoute relies on in-memory state instead of kernel state

isValidRoute() checks the in-memory configuredFwmarks map first — if the entry does not exist there, it returns false immediately without querying the kernel. After a container 
restart, the map is empty (process state is lost), but the network namespace persists (Pod is not recreated). So isValidRoute always returns false even when the route and rule are 
perfectly valid in the kernel. This causes unnecessary teardown/re-add cycles and, combined with problem 1, leads to the EEXIST failure when the target IP has changed.

## Impact

- NFQLB sets the fwmark on matched packets, but no ip rule fwmark X lookup Y exists to steer them
- Traffic is not forwarded to targets
- The error repeats on every reconcile — the LB never recovers without manual intervention or Pod recreation

## Reproduction

Any scenario where a fwmark's target IP changes:
- Pod replacement (new Pod gets same Maglev ID, different secondary IP)
- Target Pod secondary interface IP change (e.g., after node restart with IPAM reassignment)
- Scale-in followed by scale-out where the LB only reconciles the final state (freed fwmark reassigned to new Pod with different IP)
- DG deletion and re-creation where Pods were replaced in between, causing the fresh allocation to assign different IPs to the same fwmarks

The bug triggers regardless of whether the LB container was restarted.

## Proposed Fix

- Use netlink.RouteReplace instead of netlink.RouteAdd — atomically replaces any existing route in the table regardless of the current gateway
- Either remove isValidRoute (since RouteReplace is cheap and always correct) or rewrite it to query the kernel instead of relying on in-memory state which is lost anyways on process/container restart

ljkiraly:
Replace RouteAdd with RouteReplace to atomically overwrite stale routes when a fwmark's target IP changes (Pod replacement, container restart with kernel state surviving, DG re-creation).

Also tolerate EEXIST from RuleAdd (rule may already exist in kernel after container restart) and remove isValidRoute which relied on in-memory state lost on restart.

Introduces netlinkOps interface for testability.

Fixes: gh-129

### DIFF
--- a/internal/controller/loadbalancer/routing.go
+++ b/internal/controller/loadbalancer/routing.go
@@ -29,6 +29,21 @@ type RoutingManager struct {
 	configuredFwmarks map[int]routeInfo
 	// Allow disabling for tests
 	disabled bool
+	// Netlink operations (abstracted for testing).
+	// *netlink.Handle satisfies this interface directly.
+	nl netlinkOps
+}
+
+// netlinkOps abstracts netlink operations for testability.
+// *netlink.Handle satisfies this interface directly.
+type netlinkOps interface {
+	RuleAdd(rule *netlink.Rule) error
+	RuleDel(rule *netlink.Rule) error
+	RuleList(family int) ([]netlink.Rule, error)
+	RouteReplace(route *netlink.Route) error
+	RouteDel(route *netlink.Route) error
+	NeighList(link, family int) ([]netlink.Neigh, error)
+	NeighDel(neigh *netlink.Neigh) error
 }
 
 type routeInfo struct {
@@ -41,6 +56,7 @@ func NewRoutingManager() *RoutingManager {
 	return &RoutingManager{
 		configuredFwmarks: make(map[int]routeInfo),
 		disabled:          false,
+		nl:                &netlink.Handle{},
 	}
 }
 
@@ -49,15 +65,17 @@ func NewMockRoutingManager() *RoutingManager {
 	return &RoutingManager{
 		configuredFwmarks: make(map[int]routeInfo),
 		disabled:          true,
+		nl:                &netlink.Handle{},
 	}
 }
 
 // AddRoute configures policy routing for a target
 // Creates: ip rule add fwmark <fwmark> table <tableID>
 //
-//	ip route add default via <targetIP> table <tableID>
+//	ip route replace default via <targetIP> table <tableID>
 //
-// Kernel will determine the interface based on target IP subnet
+// Uses RouteReplace to atomically handle stale routes (e.g., target IP change,
+// container restart with kernel state surviving).
 func (r *RoutingManager) AddRoute(fwmark int, targetIP string) error {
 	// Skip if disabled (for tests)
 	if r.disabled {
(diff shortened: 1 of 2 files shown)


internal/controller/loadbalancer/routing.go
```diff
@@ -124,41 +133,35 @@ func (r *RoutingManager) AddRoute(fwmark int, targetIP string) error {
...
-	}
-
-	route := &netlink.Route{
-		Gw:    ip,
-		Table: info.tableID,
+// ensureRule adds the desired rule only if it doesn't already exist.
+// Linux allows duplicate ip rules, so we check first to avoid accumulating duplicates
+// across reconciles.
+func (r *RoutingManager) ensureRule(desired *netlink.Rule) error {
+	rules, err := r.nl.RuleList(desired.Family)
```
ljkiraly, line 142: No 'RuleGet' just 'RuleListFiltered', which fetches all rules from the kernel via netlink dump and filters in userspace (same as RuleList + manual filter).

ljkiraly (commented):
Verified

internal/controller/loadbalancer/routing.go
```diff
@@ -94,27 +100,30 @@ func (r *RoutingManager) AddRoute(fwmark int, targetIP string) error {
...
 	route := &netlink.Route{
 		Gw:    ip,
 		Table: tableID,
 	}
 
-	if err := netlink.RouteAdd(route); err != nil {
-		// Cleanup rule on failure
-		_ = netlink.RuleDel(rule)
-		return fmt.Errorf("failed to add route for fwmark %d: %w", fwmark, err)
+	if err := r.nl.RouteReplace(route); err != nil {
```
zolug, line 125: Would it make sense to remove the associated rule if RouteReplace failed (similarly to how the old code behaved)?

zolug (approved):

internal/controller/loadbalancer/routing.go
ljkiraly: Yes, the deleted part should remain:
```
		// Cleanup rule on failure
		_ = netlink.RuleDel(rule)
```

PR #139, opened 2026-05-13
LB controller: propagate NFQLB failure to process lifecycle

### LINKED ISSUE #135: LB controller: propagate NFQLB failure to process lifecycle
### Problem

If the NFQLB process fails to start or crashes at runtime, the LB controller continues running with healthy probes. The ENC controller includes the LB Pod in next-hop lists (it appears healthy), but traffic is black-holed because no nfqueue processor exists.

Same class of problem as limitation no.12 (BIRD error propagation in the router).

### Current Behavior

- `lbFactory.Start(ctx)` runs in a goroutine, logs errors, returns silently
- LB controller keeps reconciling (creating routes/rules/nftables) for a dead data plane
- Health/readiness probes remain green
- ENC controller routes application traffic to this Pod → black hole

### Proposed Fix

**NFQLB failure → context cancellation (process exit):**

Use `errCh` pattern: NFQLB goroutine sends error to channel, main select cancels context → process exits → Kubernetes restarts container via CrashLoopBackOff.

```go
errCh := make(chan error, 1)
go func() {
    if err := lbFactory.Start(ctx); err != nil {
        errCh <- fmt.Errorf("NFQLB process failed: %w", err)
    }
}()

// In main select or via context cancellation
select {
case err := <-errCh:
    cancel()
    return err
case <-ctx.Done():
    return nil
}
```

**Routing errors → fail health probes:**

Errors in `reconcileTargets` and `reconcileFlows` from NFQLB operations (AddTarget, DeleteTarget, flow configuration) indicate the data plane is non-functional. These should cause the readiness probe to fail, signaling to the ENC controller (via future readiness gates or container readiness) that this LB Pod should be excluded from next-hop lists.

### Impact

Without this fix, a crashed NFQLB process causes silent traffic loss with no recovery path other than manual Pod deletion.

### Related

- Limitation no.12  (BIRD error propagation in router) - same pattern

ljkiraly:
Use WithCancelCause to cancel the manager context when the NFQLB process fails. This causes the LB container to exit, triggering Kubernetes CrashLoopBackOff restart. Without this, a dead NFQLB process leaves the LB Pod appearing healthy while black-holing traffic.

Fixes: gh-135

### DIFF
--- a/cmd/stateless-load-balancer/cmd/run.go
+++ b/cmd/stateless-load-balancer/cmd/run.go
@@ -66,7 +66,7 @@ func newCmdRun() *cobra.Command {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runLoadBalancer(cmd.Context(), cfg)
+			return runLoadBalancer(cfg)
 		},
 	}
 
@@ -75,7 +75,7 @@ func newCmdRun() *cobra.Command {
 	return cmd
 }
 
-func runLoadBalancer(ctx context.Context, cfg *config.LoadBalancerConfig) error {
+func runLoadBalancer(cfg *config.LoadBalancerConfig) error {
 	setupLog.Info("Starting LoadBalancer controller", "config", cfg)
 	var tlsOpts []func(*tls.Config)
 	if !cfg.EnableHTTP2 {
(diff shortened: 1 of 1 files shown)


internal/controller/loadbalancer/targets.go
```diff
@@ -163,5 +169,5 @@ func (c *Controller) reconcileTargets(ctx context.Context, distGroup *meridio2v1
 	}
 
 	logr.Info("Reconciled targets", "distGroup", distGroup.Name, "count", len(newTargets))
-	return nil
+	return errFinal
```
zolug: I hope false-positive errors cannot surface (due to the error concatenation) that would force the controller into a vicious cycle.

zolug (approved):

cmd/stateless-load-balancer/cmd/run.go
```diff
@@ -93,12 +93,19 @@ func runLoadBalancer(ctx context.Context, cfg *config.LoadBalancerConfig) error
...
+	// would black-hole traffic since the ENC controller includes it in next-hop lists.
+	// ctrl.SetupSignalHandler() provides OS signal handling (SIGTERM/SIGINT).
+	ctx, cancel := context.WithCancelCause(ctrl.SetupSignalHandler())
+	defer cancel(nil)
+
 	go func() {
 		setupLog.Info("Starting NFQLB process")
 		if err := lbFactory.Start(ctx); err != nil {
 			setupLog.Error(err, "NFQLB process failed")
+			cancel(fmt.Errorf("NFQLB process failed: %w", err))
```
zolug, line 108: On graceful shutdown seemingly this would still log an error, which wouldn't do any harm, but might be confusing.
And the removed info log could have been useful for correlating with other events in the logs e.g. during troubleshooting.

so maybe sg like this could have been better:
```go
go func() {
    setupLog.Info("Starting NFQLB process")
    err := lbFactory.Start(ctx)
    if err != nil && ctx.Err() == nil {
        setupLog.Error(err, "NFQLB process failed")
        cancel(fmt.Errorf("NFQLB process failed: %w", err))
    } else {
        setupLog.Info("NFQLB process terminated", "error", err) // The "error" logs nil during clean exit and logs the actual error during shutdown-with-error. Operators investigating a crash-during-shutdown would see it.
    }
}()


```

internal/controller/loadbalancer/targets.go
ljkiraly: Code for error collection and handling moved to the refactor PR: #109

PR #142, opened 2026-05-14
Router controller: add connectivity classification and gate discovery

ljkiraly:
Implement first steps of gh-122 (Router connectivity status exposure):
- classifyConnectivityByFamily: determines per-IP-family connectivity from BIRD protocol statuses and a protocol-to-family mapping
- buildFamilyMap: maps NBR-<name> protocol names to IP families based on GatewayRouter spec.address
- ConnectivityGateManager with DiscoverGates: reads Pod readinessGates to determine which connectivity gates to manage

Ref: gh-122

### DIFF
--- a/cmd/router/cmd/run.go
+++ b/cmd/router/cmd/run.go
@@ -20,15 +20,16 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"time"
 
 	"github.com/spf13/cobra"
 	"golang.org/x/sync/errgroup"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
@@ -175,7 +176,7 @@ func runRouter(ctx context.Context, cfg *config.RouterConfig) error {
 	})
 
 	g.Go(func() error {
-		return monitorConnectivity(ctx, mgr, birdInstance, cfg.MonitorInterval)
+		return monitorConnectivity(ctx, mgr, birdInstance, cfg)
 	})
 
 	g.Go(func() error {
@@ -190,8 +191,10 @@ func runRouter(ctx context.Context, cfg *config.RouterConfig) error {
 	return nil
 }
 
-// monitorConnectivity monitors BGP connectivity and logs status changes
-func monitorConnectivity(ctx context.Context, mgr ctrl.Manager, birdInstance *bird.Bird, interval time.Duration) error {
+// monitorConnectivity monitors BGP connectivity and manages readiness gates
+func monitorConnectivity(
+	ctx context.Context, mgr ctrl.Manager, birdInstance *bird.Bird, cfg *config.RouterConfig,
+) error {
 	log := ctrl.Log.WithName("monitor")
 
 	// Wait for manager cache to sync
@@ -199,12 +202,34 @@ func monitorConnectivity(ctx context.Context, mgr ctrl.Manager, birdInstance *bi
 		return fmt.Errorf("failed to wait for cache sync")
 	}
 
+	// Initialize connectivity gate manager (if Pod identity is configured)
+	var gateMgr *router.ConnectivityGateManager
+	if cfg.PodName != "" && cfg.PodNamespace != "" {
+		gateMgr = router.NewConnectivityGateManager(
+			mgr.GetClient(), cfg.PodName, cfg.PodNamespace,
+			types.UID(cfg.PodUID), cfg.ConnectivityHoldTime,
+		)
+		if err := gateMgr.DiscoverGates(ctx); err != nil {
+			log.Error(err, "failed to discover readiness gates, continuing without gate management")
+			gateMgr = nil
+		} else if gateMgr.HasIPv4Gate() || gateMgr.HasIPv6Gate() {
+			log.Info("readiness gates discovered", "ipv4", gateMgr.HasIPv4Gate(), "ipv6", gateMgr.HasIPv6Gate())
+			if err := gateMgr.SetAllGatesFalse(ctx); err != nil {
+				log.Error(err, "failed to set gates to False on startup")
+			}
+		} else {
+			gateMgr = nil // No gates declared, skip gate management
+		}
+	}
+
 	// Start monitoring with configured interval
-	statusCh, err := birdInstance.Monitor(ctx, interval)
+	statusCh, err := birdInstance.Monitor(ctx, cfg.MonitorInterval)
 	if err != nil {
 		return fmt.Errorf("failed to start monitoring: %w", err)
 	}
 
+	// Build family map from GatewayRouters (refreshed on each status update)
+	var familyMap map[string]string
 	var lastCount int
 	firstUpdate := true
 
@@ -229,12 +254,50 @@ func monitorConnectivity(ctx context.Context, mgr ctrl.Manager, birdInstance *bi
 				} else {
 					log.Info("Gateway connectivity lost", "status", status.StatusString())
 				}
-
 				lastCount = count
 				firstUpdate = false
 			}
+
+			// Update readiness gates
+			if gateMgr != nil {
+				// Rebuild family map if nil or if protocols exist but map yields no matches
+				// (handles late GatewayRouter creation)
+				if familyMap == nil || (len(status.Protocols) > 0 && len(familyMap) == 0) {
+					routers, err := getGatewayRoutersFromCache(ctx, mgr.GetClient(), cfg.GatewayName, cfg.GatewayNamespace)
+					if err == nil && len(routers) > 0 {
+						familyMap = router.BuildFamilyMap(routers)
+					}
+				}
+				if familyMap != nil {
+					ipv4, ipv6 := router.ClassifyConnectivityByFamily(status.Protocols, familyMap)
+					if err := gateMgr.OnStatusUpdate(ctx, ipv4, ipv6); err != nil {
+						log.Error(err, "failed to update readiness gates")
+					}
+				}
+			}
 		}
 	}
+}
+
+func getGatewayRoutersFromCache(
+	ctx context.Context, c client.Client, gatewayName, gatewayNamespace string,
+) ([]*meridio2v1alpha1.GatewayRouter, error) {
+	list := &meridio2v1alpha1.GatewayRouterList{}
+	if err := c.List(ctx, list, client.InNamespace(gatewayNamespace)); err != nil {
+		return nil, err
+	}
+	var result []*meridio2v1alpha1.GatewayRouter
+	for i := range list.Items {
+		ref := list.Items[i].Spec.GatewayRef
+		ns := list.Items[i].Namespace
+		if ref.Namespace != nil {
+			ns = string(*ref.Namespace)
+		}
+		if string(ref.Name) == gatewayName && ns == gatewayNamespace {
+			result = append(result, &list.Items[i])
+		}
+	}
+	return result, nil
 }
 
 func protocolsUp(protocols []bird.ProtocolStatus) int {
--- a/cmd/router/cmd/run.go
+++ b/cmd/router/cmd/run.go
@@ -262,8 +262,12 @@ func monitorConnectivity(
 				// Rebuild family map on every tick — reads from informer cache (no API call)
 				// and BuildFamilyMap is a trivial loop. Handles GatewayRouter add/remove/replace.
 				routers, err := getGatewayRoutersFromCache(ctx, mgr.GetClient(), cfg.GatewayName, cfg.GatewayNamespace)
-				if err == nil && len(routers) > 0 {
-					ipv4, ipv6 := router.ClassifyConnectivityByFamily(status.Protocols, router.BuildFamilyMap(routers))
+				if err == nil {
+					var ipv4, ipv6 bool
+					if len(routers) > 0 {
+						ipv4, ipv6 = router.ClassifyConnectivityByFamily(status.Protocols, router.BuildFamilyMap(routers))
+					}
+					// When no routers exist, ipv4/ipv6 stay false — gates set to False
 					if err := gateMgr.OnStatusUpdate(ctx, ipv4, ipv6); err != nil {
 						log.Error(err, "failed to update readiness gates")
 					}
--- a/cmd/router/cmd/run.go
+++ b/cmd/router/cmd/run.go
@@ -229,7 +229,6 @@ func monitorConnectivity(
 	}
 
 	// Build family map from GatewayRouters (refreshed on each status update)
-	var familyMap map[string]string
 	var lastCount int
 	firstUpdate := true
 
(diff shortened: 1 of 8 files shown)


bszirtes (approved):
### Specific Code Issues

#### 🟡 Medium Priority Issues

**1. Family map detection could be more robust**
- **File**: `internal/controller/router/connectivity.go:128`
- **Issue**: `strings.Contains(address, ":")` will incorrectly classify IPv4-mapped IPv6 addresses
- **Fix**: Use `net.ParseIP()` and `ip.To4()` for robust detection
- **Impact**: Low (IPv4-mapped IPv6 addresses are rare in BGP peering)

**2. Gate condition patching doesn't handle conflicts**
- **File**: `internal/controller/router/connectivity.go:145`
- **Issue**: `Status().Update()` can fail with conflict error, not handled by caller
- **Fix**: Check `IsConflict()` in `patchGateCondition()` and `OnStatusUpdate()`
- **Impact**: Low (conflicts are rare, 1s monitor interval provides natural retry)

**3. Monitor loop error handling could be more granular**
- **File**: `cmd/router/cmd/run.go:238`
- **Issue**: `OnStatusUpdate()` errors are logged but don't distinguish transient vs permanent errors
- **Fix**: Check `IsConflict()` and `IsForbidden()` to handle RBAC errors gracefully
- **Impact**: Low (RBAC errors are rare, logging is acceptable)

#### 🟢 Low Priority Issues

**1. Family map refresh could be more efficient**
- **File**: `cmd/router/cmd/run.go:230`
- **Issue**: If GatewayRouters are never created, `getGatewayRoutersFromCache()` is called every second
- **Fix**: Add rate-limiting (e.g., refresh every 10 seconds)
- **Impact**: Very low (API call is cached)

**2. No validation for hold time**
- **File**: `internal/common/config/router.go:47`
- **Issue**: `ConnectivityHoldTime` can be set to negative values
- **Fix**: Add validation in `runRouter()` to reject negative values
- **Impact**: Very low (negative values are unlikely)

**3. Time-based tests use real sleep**
- **File**: `internal/controller/router/connectivity_test.go:400`
- **Issue**: Tests use `time.Sleep()`, making them slow
- **Fix**: Use a fake clock interface for faster tests
- **Impact**: Very low (tests are still fast enough for CI)

**4. No package-level comment**
- **File**: `internal/controller/router/connectivity.go:17`
- **Issue**: Missing package-level comment explaining the connectivity subpackage
- **Fix**: Add package comment
- **Impact**: Very low (code is well-documented otherwise)

**5. No Pod template example in docs**
- **File**: `docs/controllers/router.md:129`
- **Issue**: Missing example showing how to declare readiness gates in Pod template
- **Fix**: Add YAML example
- **Impact**: Very low (docs are excellent otherwise)

internal/controller/router/constants.go
```diff
@@ -0,0 +1,6 @@
+package router
+
+const (
+	ReadinessGateIPv4 = "meridio-2.nordix.org/ipv4-connectivity"
```
zolug, line 4: I think these should be in a common pkg shared by the gateway and router controllers.

cmd/router/cmd/run.go
```diff
@@ -219,12 +245,50 @@ func monitorConnectivity(ctx context.Context, mgr ctrl.Manager, birdInstance *bi
...
-
 				lastCount = count
 				firstUpdate = false
 			}
+
+			// Update readiness gates
+			if gateMgr != nil {
+				// Rebuild family map if nil or if protocols exist but map yields no matches
+				// (handles late GatewayRouter creation)
+				if familyMap == nil || (len(status.Protocols) > 0 && len(familyMap) == 0) {
```
zolug: The familyMap rebuild logic in monitorConnectivity has a gap. The rebuild condition `(familyMap == nil || (len(status.Protocols) > 0 && len(familyMap) == 0))` only handles late GatewayRouter creation from scratch.
But it won't catch runtime GatewayRouter replacements, address changes, or add/remove operations that keep the map non-empty.
Since getGatewayRoutersFromCache reads from the informer cache (no API call) and BuildFamilyMap is a trivial loop, just rebuild on every tick unconditionally, no need for a staleness heuristic.

internal/controller/router/connectivity.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	return nil
+}
+
+// OnStatusUpdate handles per-IP-family connectivity changes with damping.
+// Down transitions are immediate. Up transitions require holdTime to elapse
+// with continuous connectivity before the gate is set to True.
+func (cgm *ConnectivityGateManager) OnStatusUpdate(ctx context.Context, ipv4Connected, ipv6Connected bool) error {
+	if err := cgm.handleGate(ctx, cgm.ipv4Gate, ipv4Connected, &cgm.ipv4UpSince, ReadinessGateIPv4); err != nil {
+		return err
```
zolug: The families are independent. An API error for one shouldn't block the other IMHO.

internal/controller/router/connectivity.go
```diff
@@ -0,0 +1,235 @@
...
+			Status:             newStatus,
+			LastTransitionTime: metav1.Now(),
+		})
+	}
+
+	return cgm.client.Status().Update(ctx, pod)
+}
+
+// SetAllGatesFalse(ctx) error — called on startup (defense-in-depth)
+func (cgm *ConnectivityGateManager) SetAllGatesFalse(ctx context.Context) error {
```
zolug, line 162: Same issue as with OnStatusUpdate(). It bails on first error, potentially leaving one gate unset. Should attempt both and join errors.

internal/controller/router/connectivity.go
```diff
@@ -0,0 +1,235 @@
...
+// Down transitions are immediate. Up transitions require holdTime to elapse
+// with continuous connectivity before the gate is set to True.
+func (cgm *ConnectivityGateManager) OnStatusUpdate(ctx context.Context, ipv4Connected, ipv6Connected bool) error {
+	if err := cgm.handleGate(ctx, cgm.ipv4Gate, ipv4Connected, &cgm.ipv4UpSince, ReadinessGateIPv4); err != nil {
+		return err
+	}
+	return cgm.handleGate(ctx, cgm.ipv6Gate, ipv6Connected, &cgm.ipv6UpSince, ReadinessGateIPv6)
+}
+
+func (cgm *ConnectivityGateManager) handleGate(ctx context.Context, gate *bool, connected bool, upSince *time.Time, conditionType string) error {
```
zolug, line 191: **IMHO the correct pattern considering the cache would be:**

1. In-memory state + dirty flag is the sole decision-maker for whether to write:
   - Skip write only when desiredState == *gate && !dirty
   - On successful write: *gate = desiredState, dirty = false
   - On failed write: leave *gate unchanged, set dirty = true → forces unconditional write on next tick regardless of whether desired state matches in-memory

2. Unconditional write function: uses Status().Patch() (SSA or MergePatch), no Get from cache, no "already correct?" check inside the write function. The caller decides whether to call it, the write function always writes.

3. The cache reads are only safe for initial discovery: the first Get through the manager client triggers a fresh List from the API server (informer startup), so DiscoverGates at startup reads accurate data. Subsequent reads from the cache may be stale due to the asynchronous watch propagation delay between API server writes and informer cache updates. Basing write-or-skip decisions on cached reads can lead to:
   - Skipping a needed write (cache shows desired value but actual state differs)
   - Writing with a stale resourceVersion → 409 Conflict (when using Update)

4. Both families handled independently — OnStatusUpdate and SetAllGatesFalse should attempt both IPv4 and IPv6 operations regardless of individual failures (errors.Join), since the families are independent.

**Failed write can imply:**

Not all errors are equal, the outcome depends on the failure type.

Deterministic failures (write did NOT apply):
   - 409 Conflict — resourceVersion mismatch, server rejected the write, old state preserved
   - 404 Not Found — Pod doesn't exist
   - 403 Forbidden — RBAC denied
   - 422 Unprocessable Entity — validation rejected

For these, the state on the API server is unchanged. Retrying with fresh data will succeed. The most realistic failure in this category is 409 Conflict — kubelet frequently updates Pod status (container states, readiness probes), and if it writes between the router's Get and Update, the resourceVersion becomes stale.

Ambiguous failures (outcome unknown):
   - Connection reset / TCP RST mid-request — the request was sent but the response was lost; the API server may or may not have persisted the change
   - API server crash after accepting the write but before responding

For ambiguous failures, the caller cannot know whether the gate condition was updated. However, in practice these are rare within a cluster: the controller and API server communicate over stable local networking, and the persistent HTTP/2 connection is reused across calls. A connection reset mid-request requires the TCP connection to die between request send and response receive — a narrow window. API server rolling upgrades or node network events are the most plausible triggers.

Since Status().Patch() with the same desired value is idempotent, retrying is always safe regardless of the actual outcome.

**Problems with the current approach in the PR:**

1. patchGateCondition does Get + Update (read-modify-write) with a cache-based short-circuit:
   - The Get reads from cache, which may be stale (watch propagation delay after a recent successful write — controller-runtime's cache is not write-through)
   - The "no change needed" check (if condition.Status == newStatus { return nil }) relies on this potentially stale data
   - This can cause state drift with no error involved — a successful write followed by a quick state change can be silently skipped:
     1. Gate False on API. Connectivity up → patchGateCondition(true) → succeeds → API now True, *gate = true
     2. Connectivity drops immediately → handleGate calls patchGateCondition(false)
     3. Get from cache → cache hasn't synced the True write from step 1 → sees False → "no change needed" → returns nil without writing
     4. *gate = false (caller assumes success). API server has True. Gate stuck True while connectivity is down.
   - The root cause: the cache-based check inside the write function contradicts the caller's authoritative in-memory state. A fast state transition (True → False within the cache sync window) is silently dropped.
   - Additionally, ambiguous write failures (connection reset after API server persisted the change) can cause similar drift: in-memory is not updated, but the API state changed — if conditions then match in-memory, the needed correction is skipped.
   - Update requires the full object with correct resourceVersion → susceptible to 409 Conflict with kubelet, adding retry latency

2. No dirty flag — ambiguous failures compound the drift risk:
   - On write failure, *gate is not updated (correct)
   - For deterministic failures (409), self-healing works on next tick (desired still differs from in-memory)
   - For ambiguous failures, the write may have applied but in-memory doesn't reflect it. If input conditions then change so that desiredState == *gate, the write is skipped — leaving the API in the wrong state
   - The dirty flag pattern eliminates this at negligible cost by forcing an unconditional write after any failure

3. Early return on first family failure:
   - OnStatusUpdate returns immediately if IPv4 handleGate fails, skipping IPv6 entirely
   - SetAllGatesFalse same pattern — first failure aborts second family
   - The families are independent; one failing should not block the other

3. Recommendation:
   - Replace Status().Update() with Status().Patch() — eliminates the need for Get, avoids 409 conflicts with kubelet
   - Remove the cache-based "no change needed" check from the write function — let the caller's in-memory state + dirty flag be the sole gatekeeper
   - Add a per-family dirty flag to force retry after any failure

4. Recommendation:
   - Replace Status().Update() with Status().Patch() — eliminates the need for Get, avoids 409 conflicts with kubelet, reduces API calls from 2 to 1 per transition
   - **Remove the cache-based "no change needed" check entirely** from the write function — this is the primary source of state drift. Let the caller's in-memory state be the sole gatekeeper for whether to call the write function
   - Add a per-family dirty flag to force retry after any failure (defends against both the cache-sync race and rare ambiguous failures)
   - Use errors.Join in OnStatusUpdate and SetAllGatesFalse to handle both families independently

internal/common/config/router.go
```diff
@@ -46,6 +51,14 @@ func (c *RouterConfig) AddFlags(fs *pflag.FlagSet) {
 		"Name of the Gateway resource to watch (required)")
 	fs.StringVar(&c.GatewayNamespace, "gateway-namespace", "default",
 		"Namespace of the Gateway resource")
+	fs.StringVar(&c.PodName, "pod-name", "",
+		"Pod name (from Downward API)")
+	fs.StringVar(&c.PodNamespace, "pod-namespace", "",
+		"Pod namespace (from Downward API)")
+	fs.StringVar(&c.PodUID, "pod-uid", "",
+		"Pod UID (from Downward API)")
+	fs.DurationVar(&c.ConnectivityHoldTime, "connectivity-hold-time", 10*time.Second,
```
zolug: 10s default means every Pod in a rolling update adds 10s to the rollout (BGP establish time + 10s hold + probe interval). For a 4-replica Deployment with maxUnavailable: 1, that's ~40s+ added purely from damping.
A 3s hold (matching the BGP hold time minimum) would be more reasonable as a default — long enough to filter out a single flap, short enough to not dominate upgrade time.
The user can increase it if their environment has flappy peers. But the default shouldn't penalize the common case (stable local BGP peering).

ljkiraly: Fixed by 8405334

internal/controller/router/connectivity.go
ljkiraly: Solved by 3ff4d84

internal/controller/router/connectivity.go
ljkiraly: Fixed by ccbd2f4

internal/controller/router/connectivity.go
ljkiraly: Fixed by ccbd2f4

cmd/router/cmd/run.go
ljkiraly: Solved by 36f6013

cmd/router/cmd/run.go
```diff
@@ -221,12 +246,45 @@ func monitorConnectivity(ctx context.Context, mgr ctrl.Manager, birdInstance *bi
...
 				lastCount = count
 				firstUpdate = false
 			}
+
+			// Update readiness gates
+			if gateMgr != nil {
+				// Rebuild family map on every tick — reads from informer cache (no API call)
+				// and BuildFamilyMap is a trivial loop. Handles GatewayRouter add/remove/replace.
+				routers, err := getGatewayRoutersFromCache(ctx, mgr.GetClient(), cfg.GatewayName, cfg.GatewayNamespace)
+				if err == nil && len(routers) > 0 {
```
zolug: Misses event when all GatewayRouters are deleted. The gates retain whatever state they had before the 
deletion. If they were True (connectivity was up), they stay True indefinitely even though there are now no BGP peers and no external connectivity.

internal/controller/router/connectivity.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	// Connected
+	if *gate && !*dirty {
+		return nil // Already True, no pending write
+	}
+
+	// Start hold timer if not already started
+	now := time.Now()
+	if upSince.IsZero() {
```
zolug, line 225: This might be nitpicking, but the current damping logic's principle is: "after any down event, require the hold time to prove stability before declaring connectivity up."

Yet, in the following scenario:
1. Gate is True
2. Down event → write False fails → dirty=true, gate stays True on API
3. Up again → hold timer starts... but gate is still True on the API server the entire time

The hold timer runs, but it has no effect — the gate never went False, so there's nothing to "hold before setting True." The LB is announced as viable the entire time, including during the instability window (the latter is unavoidable).

The inconsistency: If the write had succeeded in step 2, the gate would be False for the full hold duration before going True again. But because it failed, the gate stays True — bypassing the damping entirely.

By principle, IMHO the correct behavior would be: After a failed False write with dirty=true, if connectivity recovers, the code should still force-write False first. Thus ensuring that the "cautious" path is always taken after instability, regardless of write success.

Documenting this loophole might be sufficient though. The scenario requires a failed API write (rare) followed by connectivity recovery within 1 tick (narrow window). The practical impact is: one missed damping cycle during a very specific failure+timing combination. The LB still has connectivity (it recovered), so traffic isn't actually broken — it's just that the damping principle is violated. If connectivity drops again after the recovery, the down path will fire again and attempt to write False.

But feel free to decide on this.

cmd/router/cmd/run.go
ljkiraly: Fixed in commit f95fed8

internal/controller/router/connectivity.go
ljkiraly: Comment added by f95fed8

zolug (approved):
Looks good to me.
I guess the shared pkg to contain the common constants (gates) can be addressed in the remaining issue.

Issue #128, opened 2026-05-07
[STUDY] Investigate IP family mode management and LB dual-stack routing behavior


zolug:
## Problem

The NFQLB hash table is IP-version-agnostic — it assigns fwmarks to packets regardless of whether they are IPv4 or IPv6. Each fwmark maps to a routing table that should contain routes for the target Pod. When a target has routes for only one IP family but traffic for both families arrives (via nftables VIP sets), packets for the missing family get misrouted — they fall through to the main routing table's default route instead of being explicitly dropped.

More broadly, if the underlying OS or network namespace does not support dual-stack, configuring nftables rules, routing tables, VIPs, GatewayRouter sessions, and blackhole routes for both IPv4 and IPv6 by default may be problematic. The system needs to know what IP families it should operate with.

This raises questions about:
1. How the LB should handle dual-stack routing (blackhole safety net for missing routes)
2. Whether and how to support on-the-fly transitions between IP family modes (e.g., IPv4-only → dual-stack)
3. What configuration controls are needed and where they should live

## Context

The LB creates per-target routing tables with routes for whatever IPs are present in the endpoint data. If a target has only IPv4 but IPv6 traffic is fwmarked to its table (because IPv6 VIPs exist in nftables), the packet has no matching route and may leak to the main table.

### Blackhole safety net

A low-priority blackhole rule for fwmarked traffic would catch any packet that does not match a target route — explicit drop, no misrouting. This needs to be set up based on what IP families the LB infrastructure supports.

### Ideas discussed: Two levels of IP mode

The following is a preliminary idea, not a decided approach:

1. **Cluster/LB IP mode** (LB startup config, static): What the LB infrastructure supports. Determines which blackhole rules and nftables sets to create. Changing requires LB redeployment.
   - `DualStack` → blackholes and nftables for both families (recommended default for flexibility)
   - `IPv4` / `IPv6` → only that family's infrastructure

2. **Gateway IP mode** (GatewayConfiguration field, dynamic): What this specific Gateway operates in. Determines what the DG controller expects from Pods (which IPs must be present). Can change at runtime without LB redeployment if cluster mode is DualStack (blackholes already cover both families).

### Signalling IP mode to LBs

Several approaches were discussed for how LBs learn the expected IP families:

- **LB startup config** (static flag injected by CM at deployment time): Simple, no runtime resolution needed. Changing requires redeployment.
[...]

PR #143, opened 2026-05-14
Fix Maglev ID allocation: shared per-DG per-Gateway

### LINKED ISSUE #70: Maglev ID allocation must be shared within a DistributionGroup
### Problem

The DistributionGroup controller currently assigns Maglev IDs independently per network/IP family. Each call to assignMaglevIDs operates on a per-CIDR subset of Pods, meaning the same Pod can receive different Maglev IDs in the IPv4 and IPv6 EndpointSlices for the same DistributionGroup. Furthermore, different Pods can be assigned the same Maglev ID in different EndpointSlices.

This breaks the LoadBalancer controller, which creates a single NFQLB shared-memory hash table per DistributionGroup. The hash table maps identifiers to fwmarks and routing tables. 
When the LB controller iterates EndpointSlices for a DG, it builds a target map keyed by identifier (newTargets[identifier] = endpoint.Addresses). Since this is a plain map assignment,
when the same identifier appears in multiple slices (IPv4 and IPv6), the last-processed slice overwrites the previous entry. Routing then uses only ips[0], so only one IP family gets 
a route. This leads to all sorts of problem such as:
- **Non-deterministic routing setup**: Race of endpointSlice updates. Order of endpointSlice items after list operation.
- **Cross-Pod identifier collision**: When different Pods are assigned the same identifier in different EndpointSlices (e.g., Pod-A gets maglev:3 in IPv4, Pod-B gets maglev:3 in IPv6), one Pod's route overwrites the other. Traffic for both IP families is routed to whichever Pod was processed last.
- **Broken target removal**: Deactivating an identifier from the hash table removes it for all IP families. If identifier 5 maps to Pod-A in IPv4 and Pod-B in IPv6, removing Pod-A deactivates identifier 5, breaking IPv6 traffic to Pod-B as well.

Creating multiple NFQLB hash tables per DG (e.g., one per IP family) is not viable. NFQLB uses user-defined selectors and priorities to steer packets to hash tables, and these selectors correspond to DG-level traffic classification criteria visible to external packets. Splitting a DG into sub-tables would require fine-grained selector criteria based on cluster-internal details (IP family, network attachment) that are not meaningful at the external traffic classification level.

How Meridio v1 solved this:

In Meridio v1, the equivalent concept (Stream) used a central NSP server for identifier allocation. The flow was:

1. A Pod requested to join a Stream via the TAP API Open call.
2. The system first verified the Pod had network connectivity with the Conduit (the infrastructure containing the Stream and LBs). If not, connectivity was established via Network 
Service Mesh.
3. Only after connectivity was confirmed did identifier allocation proceed — the Pod registered with the NSP server with all its IPs (both IPv4 and IPv6) under a single identifier.
4. The LB learned the target from NSP and created fwmark routes for each IP under the same fwmark/routing table. This works because IPv4 and IPv6 routing tables are independent in the 
Linux kernel.

The key design principle: IP connectivity is a prerequisite for identifier allocation, and one identifier covers all IP families for a given Pod.

The meridio-2.x POC followed the same principle: it built separate IPv4 and IPv6 EndpointSlices, merged them by Pod UID into a single view, assigned identifiers once on the merged set, then split back into per-family slices preserving the shared identifiers.

zolug:
## Fix Maglev ID allocation: shared per-DG per-Gateway

Fixes #70

### Summary

The DG controller previously assigned Maglev IDs independently per network/CIDR, causing the same Pod to receive different IDs in IPv4 and IPv6 EndpointSlices. This broke the LB controller which uses a single NFQLB hash table per DG.

This PR fixes the allocation side: IDs are now assigned once per Pod UID across all networks within a Gateway, ensuring consistent IDs in both IPv4 and IPv6 EndpointSlices.

Note: The LB controller does not yet fully support dual-stack routing. That work is tracked separately and depends on the ongoing LB refactor.

### Changes

DG controller:
- Introduce gatewayNetworkContext to group networks per Gateway (future multi-Gateway ready)
- Merge Pods by UID across a Gateway's CIDRs before Maglev ID assignment
- Simplify assignMaglevIDs and sortPodsByCreationTime to accept []corev1.Pod (IP not relevant for allocation)
- Add multi-Gateway enforcement: Ready=False with reason MultipleGateways if >1 accepted Gateway references the DG
- Add EndpointSlice compaction (fill earlier slices after scale-in, remove empty ones)
- Sort owned slices by name for deterministic processing (cache List order is not guaranteed)
- Use maps.Copy for existing assignment extraction
- Pre-allocate map/slice capacities

Documentation:
- Update docs/controllers/distributiongroup.md — reconciliation flow, Maglev scoping rationale, compaction, multi-GW step
- Update docs/operations/constraints-and-limitations.md — dual-stack constraint status
- Update docs/operations/troubleshooting.md — internalSubnets references, MultipleGateways reason
- Add docs/design/maglev-id-shared-allocation.md — design study covering problem space, alternatives, and explored directions

Tests:
- Shared IDs across networks (32 dual-stack Pods)
- Stability across reconciles
- Partial Pod replacement (16 of 32 replaced, surviving Pods keep IDs)
- EndpointSlice compaction and full-slice preservation
- Multi-Gateway reconciliation skip
- buildReadyCondition with MultipleGateways reason

### What's NOT in this PR
- LB dual-stack route handling (separate PR, pending LB refactor)
- DG gatekeeper (require all IPs before adding new Pods)
- IP family mode management (#128)

### DIFF
--- a/docs/controllers/distributiongroup.md
+++ b/docs/controllers/distributiongroup.md
@@ -14,7 +14,9 @@ The DistributionGroup controller manages EndpointSlices for secondary network en
 - Subnet CIDR (e.g., `192.168.100.0/24`)
 - Attachment type (`NAD` for Multus, `DRA` for Dynamic Resource Allocation)
 
-**Maglev ID**: A stable integer (0-31 by default) assigned to each Pod for consistent hashing. Stored in EndpointSlice's `zone` field as `maglev:<id>`. Scoped per DistributionGroup and per network context (a Pod may have different IDs in different DGs or networks).
+**Maglev ID**: A stable integer (0-31 by default) assigned to each Pod for consistent hashing. Stored in EndpointSlice's `zone` field as `maglev:<id>`. Scoped per DistributionGroup and per Gateway — the same Pod gets the same ID across all networks (IPv4/IPv6) within a Gateway, but may have different IDs in different DGs.
+
+**Why shared allocation across networks:** The LoadBalancer uses a single NFQLB hash table per DistributionGroup. This hash table maps identifiers to fwmarks, and fwmarks determine routing tables. If different networks assigned different IDs to the same Pod, the single hash table would have conflicting entries — one Pod's route would overwrite another's, causing non-deterministic routing and cross-Pod identifier collisions. Shared allocation ensures each identifier maps to exactly one Pod across all IP families, and the LB creates per-family routes under the same fwmark (IPv4 and IPv6 routing tables are independent in Linux). See [#70](https://github.com/Nordix/Meridio-2/issues/70) for full problem description and [#106](https://github.com/Nordix/Meridio-2/issues/106) for related cross-component contract considerations.
 
 ### Design Principles
 
@@ -42,7 +44,7 @@ Gateway
 └── spec.infrastructure.parametersRef → GatewayConfiguration
 
 GatewayConfiguration
-└── spec.networkSubnets → Network contexts (CIDR + attachment type)
+└── spec.internalSubnets → Network contexts (CIDR + attachment type)
 
 EndpointSlice (owned by DistributionGroup via ownerReference)
 ├── metadata.ownerReferences → DistributionGroup (controller=true)
@@ -93,14 +95,21 @@ Only process Gateways with `Accepted=True` condition set by the Gateway controll
   - Are managed by our controller (not another implementation)
 - Gateways without `Accepted=True` are ignored (no network context extracted, no EndpointSlices created)
 
-### 5. Extract Network Contexts
+### 5. Enforce Single-Gateway Restriction
+If more than one accepted Gateway references the DG (directly or via L34Routes):
+- Set `Ready=False` with reason `MultipleGateways`
+- Skip reconciliation (existing EndpointSlices are preserved/frozen)
+- The operator must resolve the conflict by removing one Gateway's reference
+
+### 6. Extract Network Contexts (per Gateway)
 For each accepted Gateway:
 - Fetch referenced GatewayConfiguration
-- Extract subnet CIDRs and attachment types from `spec.networkSubnets`
+- Extract subnet CIDRs and attachment types from `spec.internalSubnets`
 - Normalize CIDRs to canonical form (e.g., `192.168.1.5/24` → `192.168.1.0/24`)
+- Group results per Gateway (namespaced name) to scope Maglev allocation per Gateway
 
-### 6. Filter Pods by Network
-For each network context:
+### 7. Filter Pods by Network
+For each network context within the Gateway:
 - Scrape secondary IP from Pod based on attachment type:
   - **NAD**: Parse Multus `k8s.v1.cni.cncf.io/network-status` annotation
   - **DRA**: (Future implementation)
@@ -108,20 +117,28 @@ For each network context:
 - Skip Pods without IPs in the target subnet
 - Skip primary interface IPs (only secondary networks)
 
-### 7. Assign Maglev IDs (if Type=Maglev)
-**Per network context (CIDR) within this DistributionGroup:**
-- Extract existing Pod→ID mappings from current EndpointSlices for this network
+### 8. Assign Maglev IDs (if Type=Maglev)
+**Per Gateway, across all network contexts (CIDRs) within this DistributionGroup:**
+- Merge all Pods by UID across the Gateway's networks (a dual-stack Pod is counted once)
+- Extract existing Pod→ID mappings from all current EndpointSlices for this Gateway
 - Preserve existing assignments (stability)
 - Assign new IDs from available pool (0 to `maxEndpoints-1`)
 - Sort new Pods by CreationTimestamp (deterministic assignment)
 - Enforce capacity limit: exclude Pods beyond `maxEndpoints`
+- Distribute the shared IDs to per-network EndpointSlices
+- No gatekeeping: Pods are not required to have IPs in all configured Gateway subnets.
+  A Pod with only an IPv4 address gets an ID and appears in the IPv4 slice but not the IPv6 slice.
 
 **Maglev ID Scoping:**
 
-Maglev IDs are scoped **per DistributionGroup** and **per network context**. A Pod may have different IDs in different scenarios:
+Maglev IDs are scoped **per DistributionGroup** and **per Gateway**. The same Pod gets the same ID across all networks (IPv4/IPv6) within a Gateway:
 
-- **Same Pod, different networks**: Pod-A might be ID `5` in `192.168.1.0/24` (IPv4) and ID `12` in `2001:db8::/64` (IPv6)
-- **Same Pod, different DistributionGroups**: Pod-A might be ID `3` in DG-1 and ID `7` in DG-2 (even for the same network)
+- **Same Pod, same Gateway, different networks**: Pod-A has ID `5` in both `192.168.1.0/24` (IPv4) and `2001:db8::/64` (IPv6) EndpointSlices
+- **Same Pod, different DistributionGroups**: Pod-A might be ID `3` in DG-1 and ID `7` in DG-2
+
+**Immutability enforcement:**
+
+`maxEndpoints` is immutable (enforced via CEL validation). To change capacity, create a new DistributionGroup.
 
 **Why this matters for the LoadBalancer controller:**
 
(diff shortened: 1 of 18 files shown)


ljkiraly (commented):
# Preliminary (generated) review

## Summary

The PR fixes Maglev ID allocation to assign IDs once per Pod UID across all networks within a Gateway, ensuring consistent IDs in both IPv4 and IPv6 EndpointSlices. Adds single-Gateway enforcement, EndpointSlice compaction, and comprehensive tests.

## Verdict: Approve with minor suggestions

Core fix is correct, well-tested, and aligns with Meridio v1 semantics.

---

## Key Findings

### Architecture ✅

- Shared allocation mirrors Meridio v1 (one identifier per Pod, all IPs) and POC (merge → assign → split)
- Single-Gateway restriction is pragmatic and correct for current LB architecture
- `gatewayNetworkContext` type cleanly scopes allocation per Gateway

### Correctness Issues

| Issue | Severity | Recommendation | Details
|-------|----------|----------------|-------------------|
| `maps.Copy` silently overwrites inconsistent IDs from existing slices | Low | Add warning log if overwrite changes a value | Subtle correctness issue: If existing slices have inconsistent IDs for the same Pod UID (e.g., from a previous buggy version), maps.Copy silently overwrites with the last-processed value. The iteration order over slicesByNetwork map is non-deterministic, so the "winning" ID is random. The design doc acknowledges this is expected to be consistent, but there's no validation or warning log.|
| Loop over `gwContexts` only ever iterates once (single-Gateway enforced) | Cosmetic | Simplify to direct access or add comment |The `for _, gwCtx := range gwContexts` loop will only ever iterate once. This is fine for the current design but the loop structure is misleading — it suggests multi-Gateway support that doesn't exist.|
| Frozen EndpointSlices on MultipleGateways error may cause stale routing | Medium | Consider deleting owned slices instead of preserving ||

### Missing Tests

1. Single-stack Pod in dual-stack Gateway (Pod in IPv4 only)
2. Capacity exceeded with shared allocation (33 Pods, max 32 — excluded from ALL slices)
3. Multiple Gateways enforcement unit test
4. Compaction preserves IDs after scale-in

### Meridio v1 Alignment

- ✅ Core principle preserved: one ID per Pod across all IP families
- ✅ Stability via existing assignment preservation (replaces NSP keepalive)
- Gap: No collision detection (low risk — single allocator prevents collisions by design)
- Gap: LB controller must handle targets with partial IP family presence (IPv4 only)

---

## Recommendations

1. Add warning log on inconsistent existing assignments (defense-in-depth)
2. Delete EndpointSlices on MultipleGateways instead of freezing
3. Add the 4 missing test scenarios listed above
4. Document single-iteration loop intent in `calculateDesiredSlices` (Clarify the single-iteration loop in calculateDesiredSlices — either simplify to direct access or document the future extensibility intent)

bszirtes (approved):
### Issues & Suggestions

#### 🔴 Critical Issues

**None** — No blocking issues.

#### 🟡 Medium Priority Issues

**None** — No medium-priority issues.

#### 🟢 Low Priority Issues

**Issue 1: Pre-allocation capacity could be optimized**

Location: `internal/controller/distributiongroup/slices.go:127`

```go
// Current
var allPods []corev1.Pod
```

⚠️  **Problem**: No pre-allocation capacity hint. Could allocate with estimated capacity.

**Recommendation**:
```go
// Estimate capacity: sum of unique Pods across networks (upper bound)
estimatedCapacity := 0
for , podsWithIP := range podsByNetwork {
   estimatedCapacity += len(podsWithIP)
}
allPods := make([]corev1.Pod, 0, estimatedCapacity)
```

**Severity**: 🟢 Low — Minor performance optimization, not a correctness issue.

---

**Issue 2: Comment clarity in compaction logic**

Location: `internal/controller/distributiongroup/slices.go:330`

```go
// Current
// Compact: move endpoints from later slices into earlier slices with free capacity
```

⚠️  **Problem**: Comment doesn't explain *why* compaction is needed.

**Recommendation**:
```go
// Compact: move endpoints from later slices into earlier slices with free capacity.
// This prevents fragmentation after scale-in (mirrors core EndpointSlice controller behavior).
// Example: slice-0 has 50 endpoints, slice-1 has 10 → move 10 from slice-1 to slice-0, remove slice-1.
```

**Severity**: 🟢 Low — Documentation improvement, not a code issue.

---

**Issue 3: Missing test for asymmetric network presence**

⚠️  **Problem**: Edge case is documented but not tested.

**Recommendation**:
```go
func TestCalculateMaglevSlicesAsymmetricNetworkPresence(t testing.T) {
   // Pod-1 has IPv4 only, Pod-2 has IPv6 only
   // Both should get IDs, but appear in different slices
   // Verify ID slots are reserved globally (no collision)
}
```

**Severity**: 🟢 Low — Nice-to-have for confidence, edge case is well-documented.

zolug:
> Add warning log on inconsistent existing assignments (defense-in-depth)

We don't care about backward compatibility with the broken baseline (alpha phase).
Even when upgrading from a broken version, after one reconcile (indeed random order), all slices become consistent; it's a one-time migration. From that point on: all slices are consistent, order doesn't matter anymore.
There's a comment explaining this: "All slices are expected to have consistent IDs for the same Pod UID, so iteration order over slicesByNetwork does not affect the result."


> Delete EndpointSlices on MultipleGateways instead of freezing

It was a deliberate choice. Deleting would disrupt active traffic for what might be a transient misconfiguration (e.g., operator accidentally creates a second L34Route referencing the DG, notices immediately, deletes it). Preserving existing slices keeps the data plane functional while the operator resolves the conflict. The Ready=False status signals the problem clearly. This is documented in the design study and the DG controller doc.


> Document single-iteration loop intent in calculateDesiredSlices (Clarify the single-iteration loop in calculateDesiredSlices — either simplify to direct access or document the future extensibility intent)

The loop is intentionally structured for future multi-Gateway support. It's a deliberate choice to keep the per-Gateway iteration so that lifting the single-Gateway restriction later is a natural extension. The comment above the loop already explains this: "Process each Gateway independently (scopes Maglev allocation per Gateway)."

Will add comment to convey the intent.

> Add the 4 missing test scenarios listed above

Ok. (Multiple Gateways enforcement was already part of the second commit.)

arp-est (commented):
I got the following form the colleague, that seemed relevant:

 1 The bounds check issue in maglev.go
```

  How it works today

  When the DG controller reconciles, it reads existing EndpointSlices to preserve Maglev ID stability:

  // Extract existing Pod→ID from slices written in previous reconciles
  existingAssignments := make(map[string]int32)
  for _, slices := range slicesByNetwork {
      maps.Copy(existingAssignments, extractMaglevAssignments(slices))
  }

  // Preserve them unconditionally
  for i := range pods {
      uid := string(pods[i].UID)
      if id, exists := existingAssignments[uid]; exists {
          podToID[uid] = id    // ← no check that id < maxEndpoints
          usedIDs[id] = true
      }
  }

  The problem

  maxEndpoints is immutable via CEL (self == oldSelf), so under normal operation an out-of-range ID shouldn't appear. But it can happen through:

  1. Manual EndpointSlice tampering — someone edits the zone field to maglev:999
  2. Stale slices from before immutability was enforced — if the CEL rule was added after some DGs already existed with different maxEndpoints values
  3. Restore from backup — EndpointSlices restored from a backup where maxEndpoints was different

  What goes wrong downstream

  The LB controller uses the ID directly:

  index := identifier + 1   // NFQLB uses 1-based indexing
  fwmark := identifier + fwmarkOffset
  instance.Activate(index, fwmark)

  NFQLB's shared memory is allocated as M = maxEndpoints × 100 with N = maxEndpoints slots. If identifier = 999 but N = 32, the Activate(1000, ...) call writes beyond the shared memory region — undefined behavior
  (crash or silent corruption of adjacent memory).

  The fix

  One line change in assignMaglevIDs:

  if id, exists := existingAssignments[uid]; exists && id >= 0 && id < maxEndpoints {
      podToID[uid] = id
      usedIDs[id] = true
  } else {
      newPods = append(newPods, pods[i])
  }

  A Pod with an out-of-range stale ID is treated as a new Pod and gets a fresh valid ID from the available pool. One-time reshuffle for that Pod's hash bucket, but safe.
```

zolug:
> **Issue 1: Pre-allocation capacity could be optimized**
> **Issue 2: Comment clarity in compaction logic**
> **Issue 3: Missing test for asymmetric network presence**

Took care of issue 1 and 3.
Existing comment about compaction logic's motives are descriptive already:
```go
// Compact: move endpoints from later slices into earlier slices with free capacity.
// This prevents fragmentation after scale-in (mirrors core EndpointSlice controller behavior).
```

ljkiraly (approved):
Looks good.

PR #145, opened 2026-05-14
Add automatic gofmt to pre-commit hook

szvincze:
Auto-format and re-stage Go files before running make check, so formatting fixes are included in the commit automatically.

### DIFF
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -20,6 +20,13 @@ if [ -n "$KUSTOM_FILES" ]; then
     echo ""
 fi
 
+# Auto-format staged Go files
+STAGED_GO_FILES=$(git diff --cached --name-only --diff-filter=ACM | grep '\.go$')
+if [ -n "$STAGED_GO_FILES" ]; then
+    gofmt -w $STAGED_GO_FILES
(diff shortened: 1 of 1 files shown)


ljkiraly (commented):
The generated files might be excluded (e.g. `\.deepcopy\.go$` or maybe `\.gen\.go$` - not sure if we have), otherwise looks good.
Not sure, by the way. The generated files should be well formatted. But can not be edited - that's sure. :)

szvincze:
> The generated files might be excluded (e.g. `\.deepcopy\.go$` or maybe `\.gen\.go$` - not sure if we have), otherwise looks good. 
> Not sure, by the way. The generated files should be well formatted. But can not be edited - that's sure. :)

Basically, generated files are already well-formatted by their generators, and you are right that they shouldn't be manually edited. Running gofmt on them is harmless but unnecessary, it will do nothing since they're already formatted. And if someone accidentally edits one, the gofmt pass won't hide that problem.
So, in my opinion no need to exclude the generated files, thus the original grep '\.go$' is fine as-is.

ljkiraly (approved):

PR #146, opened 2026-05-16
Add --cert-wait-timeout flag to poll for certificates before starting

### LINKED ISSUE #78: Add optional cert-wait-timeout for controller-manager startup
When the controller-manager and cert-manager are deployed simultaneously (e.g., via a single Helm chart), the controller-manager crashes on startup because TLS certificates are not yet provisioned. Kubernetes restart backoff handles this, but it causes a few unnecessary restarts.

Proposal:

Add a `--cert-wait-timeout` flag (default 0 = fail immediately, preserving current behavior). When set to a positive duration, the controller-manager polls for required certificate files before starting the manager. Only wait for certs that are actually needed:
- Webhook cert: if --enable-webhooks=true and --webhook-cert-path is set
- Metrics cert: if --metrics-bind-address is not "0", --metrics-secure=true, and --metrics-cert-path is set

All files should be checked concurrently so the total wait is bounded by the timeout, not multiplied by the number of files.

szvincze:
This is a solution for #78.

When the controller-manager and cert-manager are deployed simultaneously, TLS certificates may not be provisioned yet at startup. This flag allows the controller-manager to poll for required certificate files before starting the manager.

Behavior:
- 0: no wait, fail immediately if certs are missing
- \>0: wait up to the specified duration
- Max: 1, values above are rejected at startup
- Default: 10s

Only waits for certs that are actually needed:
- Webhook cert: when `--enable-webhooks=true` and `--webhook-cert-path` is set
- Metrics cert: when `--metrics-bind-address` is not 0, `--metrics-secure=true`, and `--metrics-cert-path` is set

All files are checked concurrently so total wait is bounded by the timeout.

### DIFF
--- a/cmd/controller-manager/cmd/run.go
+++ b/cmd/controller-manager/cmd/run.go
@@ -21,7 +21,6 @@ import (
 	"crypto/tls"
 	goflag "flag"
 	"fmt"
-	"path/filepath"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -112,19 +111,17 @@ func runManager(cfg *config.ManagerConfig) error {
 
 	// Wait for certificate files if configured
 	if cfg.CertWaitTimeout > 0 {
-		var certFiles []string
-		if cfg.EnableWebhooks && cfg.WebhookCertPath != "" {
-			certFiles = append(certFiles,
-				filepath.Join(cfg.WebhookCertPath, cfg.WebhookCertName),
-				filepath.Join(cfg.WebhookCertPath, cfg.WebhookCertKey),
-			)
-		}
-		if cfg.MetricsAddr != "0" && cfg.SecureMetrics && cfg.MetricsCertPath != "" {
-			certFiles = append(certFiles,
-				filepath.Join(cfg.MetricsCertPath, cfg.MetricsCertName),
-				filepath.Join(cfg.MetricsCertPath, cfg.MetricsCertKey),
-			)
-		}
+		certFiles := (&prerequisites.CertWaitConfig{
+			EnableWebhooks:  cfg.EnableWebhooks,
+			WebhookCertPath: cfg.WebhookCertPath,
+			WebhookCertName: cfg.WebhookCertName,
+			WebhookCertKey:  cfg.WebhookCertKey,
+			MetricsAddr:     cfg.MetricsAddr,
+			SecureMetrics:   cfg.SecureMetrics,
+			MetricsCertPath: cfg.MetricsCertPath,
+			MetricsCertName: cfg.MetricsCertName,
+			MetricsCertKey:  cfg.MetricsCertKey,
+		}).CertFiles()
 		if len(certFiles) > 0 {
 			timeout := cfg.CertWaitTimeout
 			ctx, cancel := context.WithTimeout(context.Background(), timeout)
--- a/cmd/controller-manager/cmd/run.go
+++ b/cmd/controller-manager/cmd/run.go
@@ -86,6 +86,10 @@ func NewRunCmd() *cobra.Command {
 				return fmt.Errorf("LB deployment template not found: %w", err)
 			}
 
+			if cfg.CertWaitTimeout > time.Minute {
+				return fmt.Errorf("--cert-wait-timeout cannot exceed 1 minute (got %s)", cfg.CertWaitTimeout)
+			}
+
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -122,11 +126,12 @@ func runManager(cfg *config.ManagerConfig) error {
 			)
 		}
 		if len(certFiles) > 0 {
-			timeout := min(cfg.CertWaitTimeout, time.Minute)
+			timeout := cfg.CertWaitTimeout
 			ctx, cancel := context.WithTimeout(context.Background(), timeout)
-			defer cancel()
 			setupLog.Info("waiting for certificate files", "files", certFiles, "timeout", timeout)
-			if err := prerequisites.WaitForCerts(ctx, certFiles); err != nil {
+			err := prerequisites.WaitForCerts(ctx, certFiles)
+			cancel()
+			if err != nil {
 				return fmt.Errorf("certificate wait failed: %w", err)
 			}
 			setupLog.Info("all certificate files are available")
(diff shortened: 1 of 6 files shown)


internal/common/config/manager.go
```diff
@@ -95,6 +96,8 @@ func (c *ManagerConfig) AddFlags(fs *pflag.FlagSet) {
 		"ServiceAccount name for LB Deployment pods.")
 	fs.IntVar(&c.WebhookPort, "webhook-port", 9443,
 		"The port the webhook server binds to.")
+	fs.DurationVar(&c.CertWaitTimeout, "cert-wait-timeout", 10*time.Second,
+		"Duration to wait for certificate files before starting. 0 means no wait. Maximum 1 minute.")
```
ljkiraly, line 100: The `min(cfg.CertWaitTimeout, time.Minute)` cap is undocumented in the flag description. The flag help says "Maximum 1 minute" but doesn't explain what happens if you pass 2m (silently capped, not rejected).

cmd/controller-manager/cmd/run.go
```diff
@@ -102,6 +105,34 @@ func NewRunCmd() *cobra.Command {
...
+		if cfg.MetricsAddr != "0" && cfg.SecureMetrics && cfg.MetricsCertPath != "" {
+			certFiles = append(certFiles,
+				filepath.Join(cfg.MetricsCertPath, cfg.MetricsCertName),
+				filepath.Join(cfg.MetricsCertPath, cfg.MetricsCertKey),
+			)
+		}
+		if len(certFiles) > 0 {
+			timeout := min(cfg.CertWaitTimeout, time.Minute)
+			ctx, cancel := context.WithTimeout(context.Background(), timeout)
+			defer cancel()
```
ljkiraly: `defer cancel()` is called after `context.WithTimeout`. The context is only used for cert waiting but the defer runs when `runManager` returns (after entire manager lifecycle). Harmless but slightly misleading. Consider scoping more tightly or removing the defer (context is short-lived).

internal/common/prerequisites/cert_wait_test.go
ljkiraly, line 1: Missing tests (nice-to-have)

- 1-minute cap behavior
- Integration test verifying flag works in `runManager`

ljkiraly (approved):
Approving with suggestions:

1. **Minor**: Consider validating `CertWaitTimeout` > 1 minute at flag parse time with an error instead of silently capping.
2. **Minor**: Scope the `defer cancel()` more tightly to the cert-wait block.

bszirtes (commented):
#### 🔴 Critical Issues

None - all issues are documentation/clarity improvements.

---

#### 🟡 Medium Priority Issues

**1. Undocumented 1-minute cap**

**File**: `internal/common/config/manager.go:99`

```go
"Duration to wait for certificate files before starting. 0 means no wait. Maximum 1 minute."
```

**Problem**: Doesn't explain that values >1m are silently capped.

**Recommendation**:
```go
"Duration to wait for certificate files before starting. 0 means no wait. Values >1m are capped at 1 minute."
```

**Impact**: Medium - users might be confused why `--cert-wait-timeout=2m` times out after 1 minute.

---

**2. defer cancel() scope confusion**

**File**: `cmd/controller-manager/cmd/run.go:127`

```go
ctx, cancel := context.WithTimeout(context.Background(), timeout)
defer cancel()  // Runs when runManager returns (after entire manager lifecycle)
```

 **Problem**: The defer runs much later than the context is used. Harmless but misleading.

**Recommendation**: Scope the context more tightly or remove the defer.

**Impact**: Low - no functional issue, just clarity.

---

#### 🟢 Low Priority Issues

**3. Missing test coverage for 1-minute cap**

**File**: `internal/common/prerequisites/cert_wait_test.go`

 **Problem**: No test verifies the 1-minute cap behavior.

**Recommendation**: Add integration test in `cmd/controller-manager/cmd/run_test.go` (if it exists) or document as known limitation.

**Impact**: Low - the cap is applied in `runManager`, not in the `prerequisites` package.

---

**4. Hardcoded poll interval**

**File**: `internal/common/prerequisites/cert_wait.go:28`

```go
const certPollInterval = 1  time.Second
```

 **Problem**: Poll interval is hardcoded. Not configurable for testing or different scenarios.

**Recommendation**: Consider making it configurable.

**Impact**: Low - 1s is reasonable for production use.

---

#### Note: Loop Variable Capture

**File**: `internal/common/prerequisites/cert_wait.go:38-41`

The current code captures loop variable `f` in goroutine closures:

```go
for , f := range files {
   g.Go(func() error {
       return waitForFile(ctx, f)
   })
}
```

This pattern is **safe in Go 1.22+** (which this project uses: Go 1.25.3) due to per-iteration loop variable semantics. However, if backporting to Go <1.22, this would require shadowing: `f := f`.

internal/common/config/manager.go
szvincze: I changed the behavior to reject `--cert-wait-timeout` values exceeding 1 minute at flag parse time instead of silently capping.

cmd/controller-manager/cmd/run.go
szvincze: I scoped context cancel to immediately after WaitForCerts returns instead of deferring to end of runManager.

internal/common/prerequisites/cert_wait_test.go
szvincze: Tests added.

szvincze:
> **1. Undocumented 1-minute cap**

Behavior changed and documented to reject values exceeding 1 minute, no silent capping anymore.

> **2. defer cancel() scope confusion**

Defer removed, context canceled immediately after WaitForCerts returns.

> **3. Missing test coverage for 1-minute cap**

Refer to my answer above at 1.

> **4. Hardcoded poll interval**

Keeping as-is, because the 1s interval is a reasonable default for a startup-only operation, and tests already control timing via context deadlines. Adding configurability here would expose an implementation detail with no practical benefit.


> Note: Loop Variable Capture

Currently, in this project Go 1.25.3 is used, no plans to backporting, so there's nothing to do.

bszirtes (approved):
Looks good to me.

ljkiraly:
All my recommendations addressed correctly. The PR is ready to merge.

PR #149, opened 2026-05-19
RFC 5881 compliant BFD port range

### LINKED ISSUE #77: BIRD; IANA approved src ports for BFD
RFC 5881 Section 4 requires BFD source ports to be in the range 49152-65535

```
   BFD Control packets MUST be transmitted in UDP packets with
   destination port 3784, within an IPv4 or IPv6 packet.  The source
   port MUST be in the range 49152 through 65535.
```

Recommendation is to use `ip_local_port_range` sysctl:
`sysctl -w net.ipv4.ip_local_port_range='49152 65535'`

szvincze:
### Problem

BIRD uses the kernel's ephemeral port allocation for BFD UDP sockets. The Linux default range (32768 60999) violates RFC 5881, and strict BFD peers may reject packets with source ports below 49152.

### Alternatives considered

1. Init container in the LB Deployment template
It requires `privileged: true` for the init container, which conflicts with the restricted Pod Security Standard currently in use.

2. Pod-level securityContext.sysctls
It would require the sysctl to be listed as "safe" in the kubelet config, or the Pod Security Admission to allow "unsafe" `sysctls. ip_local_port_range` is namespaced (per-netns) so it's safe in practice, but K8s classifies it as unsafe by default.

3. Set it inside the router process at startup
`/proc/sys` is mounted read-only in containers; `NET_ADMIN` capability is not enough, it would require `SYS_ADMIN` too.

#### `sysctl` tuning via Multus NAD
Add `"net.ipv4.ip_local_port_range": "49152 65535"` to the existing `sysctl-tuning` NAD that is already attached to LB Pods via GatewayConfiguration. This applies the correct range in the Pod's network namespace without requiring code changes or additional capabilities.

It fixes #77.

### DIFF
--- a/cmd/router/cmd/run.go
+++ b/cmd/router/cmd/run.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"os"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


ljkiraly (approved):
Key trade-off vs Meridio v1: Meridio v1 uses a privileged init container (the operator generates it), making the sysctl automatic and invisible to users. Meridio-2 delegates this to the user via GatewayConfiguration's networkAttachments (referencing the tuning NAD). This is more explicit but requires the user to include the tuning NAD - if they forget, BFD silently violates the RFC.  The documentation update and constraints doc update mitigate this.
I don't like the dependency to Multus in general, but the chosen approach is correct.

Issue #94, opened 2026-04-07
Sidecar restart recovery: stale network configuration after ENC changes during downtime


zolug:
### Problem

All sidecar in-memory state (tableIDs, managedVIPs) is lost on container restart. The Linux network namespace and its configuration (VIPs on interfaces, routing rules, routing tables) survive the restart since they belong to the Pod, not the container. If the Endpoint Network Configuration (ENC) is updated while the sidecar is down, the sidecar cannot reconcile the delta between the kernel's current state and the new desired state.

See: constraints-and-limitations.md "No sidecar restart recovery"

### Failure Scenarios

Three issues arise, all rooted in the loss of the `gatewayName` → `tableID` mapping and `managedVIPs` set:

#### 1. VIP leak

`managedVIPs` is empty after restart. `syncVIPs` only removes VIPs within its managed set, so VIPs from the previous run that are no longer in the ENC spec remain on the interface indefinitely. This is particularly problematic when a Pod leaves a Gateway or VIPs move between Gateways.

#### 2. Orphaned routes on table ID shift

The `tableIDAllocator` starts fresh after restart, assigning IDs sequentially from the configured minimum. If the gateway set changed while the sidecar was down (e.g., a gateway was removed from the ENC), previously-used table IDs may not be reallocated to any gateway. These orphaned tables retain stale routes with no code path to clean them.

#### 3. Orphaned rules on table ID shift

Same root cause as orphaned routes. Source-based policy rules (ip rule from <VIP> table <tableID>) pointing to orphaned table IDs are never cleaned because `syncRules` only scans rules matching actively-allocated table IDs.

### Specific concern: VIPs moving between Gateways

When VIPs are reassigned between Gateways while the sidecar is down:
- Old VIPs remain on the interface (leak)
- Source-based routing rules for the old VIPs still point to the old Gateway's routing table
- The new Gateway's routing table may get a different table ID after restart
- Return traffic from the moved VIPs follows the old routing path (wrong next-hops)

### Concrete scenario

Before restart:
- ENC has gateways `gw-a` and `gw-b`
- Allocator assigned: `gw-a → 50000`, `gw-b → 50001`
- Kernel: table 50000 has gw-a routes/rules, table 50001 has gw-b routes/rules

After restart, `gw-a` is removed from the ENC:
- Allocator starts fresh: `gw-b → 50000` (first and only allocation)
[...]

PR #154, opened 2026-05-27
Sidecar restart recovery: emptyDir persistence + kernel VIP scanning

bszirtes:
### Summary

Implements restart recovery for the network sidecar to eliminate three known bugs that occur when the sidecar container restarts while the ENC is modified:

1. **VIP leak** - VIPs from deleted gateways remain on interfaces
2. **Orphaned routes** - Stale routes persist after table ID reallocation
3. **Orphaned policy rules** - Stale rules persist after table ID reallocation

### Solution

Hybrid approach combining two mechanisms:

- **emptyDir table ID persistence** (`/var/run/meridio/table-id-mapping.json`) - Preserves `gatewayName → tableID` mapping across container restarts, preventing table ID shifts
- **Kernel VIP scanning** - Seeds `managedVIPs` from kernel state (`/32` and `/128` addresses on secondary interfaces) on startup, enabling cleanup of stale VIPs

### Implementation

**Core changes:**
- `internal/controller/sidecar/persistence.go` - `loadMapping()`/`saveMapping()` with atomic writes
- `internal/controller/sidecar/network.go` - `scanManagedVIPs()` scans kernel interfaces
- `internal/controller/sidecar/tableid.go` - `snapshot()`/`restore()` methods for allocator
- `internal/controller/sidecar/controller.go` - Initialize state in `SetupWithManager()`, save after successful reconcile

**Testing:**
- Unit tests: 79.2% coverage, all passing
- E2E tests: Comprehensive restart recovery validation for Common/Separate App Network suites using restart gate mechanism

### Test Design

The e2e test validates all three bugs are fixed:

1. Captures initial table IDs for two gateways
2. Sets restart gate marker to pause sidecar on next restart
3. Kills sidecar container (pauses at gate)
4. Deletes one gateway while sidecar is down
5. Releases gate, allowing sidecar to start with persisted state
6. Validates:
   - Preserved gateway keeps original table ID (no shift)
   - Deleted gateway's VIP is removed (no leak)
   - Deleted gateway's policy rules are removed (no orphans)
   - Deleted gateway's routing table is cleaned (no orphans)

Test runs automatically for suites with 2+ gateways and is fully idempotent (restores deleted gateway in `AfterAll`).

### Files Changed

**Implementation:**
- `internal/controller/sidecar/persistence.go` (new)
- `internal/controller/sidecar/persistence_test.go` (new)
- `internal/controller/sidecar/controller.go`
- `internal/controller/sidecar/network.go`
- `internal/controller/sidecar/tableid.go`
- `internal/controller/sidecar/tableid_test.go`
- `internal/controller/sidecar/controller_test.go`

**E2E tests:**
- `test/e2e/e2e_suite_appnetwork_test.go`
- `test/e2e/suites/common-appnetwork/targets.yaml`
- `test/e2e/suites/separate-appnetwork/targets.yaml`

### Related Issue

#94

### DIFF
--- a/internal/controller/sidecar/persistence.go
+++ b/internal/controller/sidecar/persistence.go
@@ -17,10 +17,14 @@ limitations under the License.
 // Package sidecar implements the network sidecar controller.
 //
 // Restart Recovery:
-// The controller persists table ID mappings to /var/run/meridio/table-id-mapping.json
-// (emptyDir volume) to prevent table ID shifts across container restarts. On startup,
-// it also scans the kernel for existing VIPs (/32 and /128 addresses on secondary
-// interfaces) to enable cleanup of stale VIPs even if the mapping file is lost.
+// The controller can persist table ID mappings to an emptyDir volume to prevent
+// table ID shifts across container restarts. The path is configurable via
+// --table-id-mapping-file (default: /var/run/meridio/table-id-mapping.json).
+// Set to empty string to disable persistence.
+//
+// On startup, it also scans the kernel for existing VIPs (/32 and /128 addresses
+// on secondary interfaces) to enable cleanup of stale VIPs even if the mapping
+// file is lost or persistence is disabled.
 package sidecar
 
 import (
@@ -29,13 +33,16 @@ import (
 	"os"
 )
 
-const mappingFile = "/var/run/meridio/table-id-mapping.json"
-
 // loadMapping reads the persisted table ID mapping and seeds the allocator.
+// If path is empty, persistence is disabled and this is a no-op.
 // Returns nil if file doesn't exist (first run).
 // Returns error if file is corrupted or contains invalid data.
-func loadMapping(allocator *tableIDAllocator) error {
-	data, err := os.ReadFile(mappingFile)
+func loadMapping(allocator *tableIDAllocator, path string) error {
+	if path == "" {
+		return nil // Persistence disabled
+	}
+
+	data, err := os.ReadFile(path)
 	if os.IsNotExist(err) {
 		return nil // First run, no file yet
 	}
@@ -59,8 +66,13 @@ func loadMapping(allocator *tableIDAllocator) error {
 }
 
 // saveMapping writes the current allocator state to disk atomically.
+// If path is empty, persistence is disabled and this is a no-op.
 // Non-fatal errors are returned but should not stop reconciliation.
-func saveMapping(allocator *tableIDAllocator) error {
+func saveMapping(allocator *tableIDAllocator, path string) error {
+	if path == "" {
+		return nil // Persistence disabled
+	}
+
 	mapping := allocator.snapshot()
 	data, err := json.Marshal(mapping)
 	if err != nil {
@@ -68,12 +80,12 @@ func saveMapping(allocator *tableIDAllocator) error {
 	}
 
 	// Atomic write: write to temp file, then rename
-	tmpFile := mappingFile + ".tmp"
+	tmpFile := path + ".tmp"
 	if err := os.WriteFile(tmpFile, data, 0600); err != nil {
 		return fmt.Errorf("failed to write temp file: %w", err)
 	}
 
-	if err := os.Rename(tmpFile, mappingFile); err != nil {
+	if err := os.Rename(tmpFile, path); err != nil {
 		return fmt.Errorf("failed to rename temp file: %w", err)
 	}
 
--- a/internal/controller/sidecar/persistence.go
+++ b/internal/controller/sidecar/persistence.go
@@ -14,6 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package sidecar implements the network sidecar controller.
+//
+// Restart Recovery:
+// The controller persists table ID mappings to /var/run/meridio/table-id-mapping.json
+// (emptyDir volume) to prevent table ID shifts across container restarts. On startup,
+// it also scans the kernel for existing VIPs (/32 and /128 addresses on secondary
+// interfaces) to enable cleanup of stale VIPs even if the mapping file is lost.
 package sidecar
 
 import (
--- a/internal/controller/sidecar/persistence.go
+++ b/internal/controller/sidecar/persistence.go
@@ -1,0 +1,74 @@
+/*
+Copyright (c) 2026 OpenInfra Foundation Europe. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sidecar
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+)
+
+const mappingFile = "/var/run/meridio/table-id-mapping.json"
+
+// loadMapping reads the persisted table ID mapping and seeds the allocator.
+// Returns nil if file doesn't exist (first run).
+// Returns error if file is corrupted or contains invalid data.
+func loadMapping(allocator *tableIDAllocator) error {
+	data, err := os.ReadFile(mappingFile)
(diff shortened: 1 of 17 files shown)


zolug:
### Correctness Assessment Criteria

The solution was evaluated against the following requirements from issue #94:

1. VIP leak prevention: VIPs from deleted/modified gateways must be removed after restart
2. Orphaned route cleanup: Stale routes from shifted/removed table IDs must be cleaned
3. Orphaned rule cleanup: Stale ip rules pointing to orphaned tables must be cleaned
4. VIP movement between gateways: VIPs reassigned to a different gateway during downtime must be correctly moved (removed from old interface, added to new)
5. Table ID stability: Same gateway should get the same table ID across restarts to avoid cascading orphans
6. Atomicity: Crash at any point during recovery or normal operation should not leave the system in a permanently broken state
7. Safety: The recovery mechanism must not incorrectly remove addresses or routes that don't belong to the sidecar

### Solution Summary

Hybrid approach:
- **emptyDir file** (/var/run/meridio/table-id-mapping.json): persists gatewayName→tableID mapping only. Atomic write after each successful reconcile. Loaded on startup to seed the allocator.
- **Kernel VIP scan** (scanManagedVIPs): scans /32 and /128 addresses on all non-lo, non-eth0 interfaces. Seeds managedVIPs on startup so syncVIPs can diff and remove stale VIPs.

VIPs are NOT persisted to the file. VIP recovery relies entirely on the kernel scan.

### Evaluation Against Criteria

1. VIP leak prevention: Partially addressed ⚠️

VIPs on interfaces that are still referenced by the current ENC are properly cleaned via syncVIPs (kernel scan seeds the managed set, reconcile removes VIPs not in desired set). 
However, VIPs on interfaces exclusively owned by a gateway that was removed entirely from the ENC while the sidecar was down are NOT cleaned — syncVIPs is only called for interfaces referenced by current ENC domains. The stale gateway cleanup path (flushTable) handles routes/rules but not VIPs.

The kernel scan finds these VIPs and adds them to managedVIPs, but since no reconcile path calls syncVIPs for out-of-scope interfaces, the information is never acted upon.

2. Orphaned route cleanup: ✅

Persisted table IDs prevent shifts. If a gateway is removed during downtime, flushTable cleans its routes on first reconcile (restored mapping enables the stale-gateway diff).

3. Orphaned rule cleanup: ✅

Same mechanism as routes — flushTable removes rules for orphaned tables.

4. VIP movement between gateways: Partially addressed ⚠️

If the old gateway still exists in the ENC (without the moved VIP): syncVIPs runs for its interface and removes the stale VIP. ✅

If the old gateway is removed entirely: same gap as criterion #1 — VIPs on its exclusive interface are not cleaned. ⚠️

5. Table ID stability: ✅

emptyDir persistence ensures same gateway → same table ID. restore() validates range and duplicates. nextID advanced past highest restored value.

6. Atomicity: Acceptable with caveat

For table IDs: crash between applyState and saveMapping means the file has old mapping. On restart, allocator may assign shifted IDs. First reconcile overwrites kernel state idempotently (RouteReplace, syncRules). Self-healing, but may cause brief traffic disruption during the shift. Not permanently broken.

For VIPs: no file persistence, so no file-staleness concern. The kernel scan on restart always reflects actual kernel state. However, if VIPs were applied to an interface, the sidecar crashes, and the ENC subsequently removes that gateway before restart — the kernel scan finds the VIPs but criterion #1's gap applies (interface out of ENC scope, no cleanup path). This is not an atomicity failure per se but a consequence of the VIP leak gap.

7. Safety: ⚠️

scanManagedVIPs assumes ALL /32 and /128 addresses on secondary interfaces (non-lo, non-eth0) are sidecar-managed VIPs. The false-positive risk materializes specifically when an interface IS referenced by the current ENC AND has non-sidecar /32 addresses not in the desired set — those would be incorrectly removed by syncVIPs. Interfaces NOT in the current ENC are never processed by syncVIPs, so false positives on those are inert (never acted upon).

In practice, interfaces managed by the sidecar should not have /32 or /128 addresses from other processes simultaneously. The realistic concern is unrelated secondary interfaces being scanned unnecessarily.

### Discarded Improvement Proposals

A. VIP cleanup for removed-gateway interfaces (partially fixes criteria #1 and #4):

Persist interface associations alongside table IDs: {gatewayName: {tableID: N, interfaces: ["net1", "net2"]}}. On stale gateway cleanup, also call 
syncVIPs(nl, link, nil, managedVIPs[ifName]) for interfaces that belonged to the removed gateway. This handles the common case (gateway removed during downtime, sidecar restarts, cleanup possible from persisted info).

Note: this still doesn't cover the edge case where VIPs were applied to a new interface (crash before that interface was persisted) AND the gateway was subsequently removed from ENC. (Not to mention more exotic use cases where ENC drifts but due recurring random errors the config cannot be saved and are partially applied before crash.)

B. Reduce false-positive VIP scan risk (fixes criterion #7):

Interface filter parameter (--managed-interface-pattern): scope the VIP scan to only interfaces matching a prefix or regex pattern (e.g., net*). Operator declares which interfaces the sidecar may touch. On declared interfaces, all /32/128 addresses are assumed to be sidecar-managed. Eliminates risk from unrelated secondary interfaces.

Trade-off: makes the sidecar aggressive on declared interfaces (removes any /32/128 not in desired set). It provides the only(?) deterministic way to handle the removed-gateway VIP leak — if we own the interface, we can clean anything on it that shouldn't be there regardless of whether the interface is in the current ENC.

Note: the interface filter is static configuration. Changing it at runtime could create its own consistency issues.

### Fundamental Limitation

VIP ownership attribution cannot be fully solved without kernel-level address tagging (which Linux doesn't provide). The sidecar cannot atomically add an address AND record ownership in a single syscall. Any approach using external persistence (file, API server, nftables maps) has a crash window between the kernel operation and the persistence write where state candiverge. The kernel scan helps but cannot distinguish "our" addresses from others' on shared interfaces.

The practical consequence: in edge cases involving crash during partial apply combined with subsequent ENC changes removing the affected config, stale VIPs may persist on out-of-scope interfaces. This is acceptable for alpha given the low probability.

### Final Summary

| Criterion | Status | Notes |
|-----------|--------|-------|
| VIP leak prevention | ⚠️ Partial | Gap on removed-gateway-exclusive interfaces |
| Orphaned routes | ✅ | Via persisted IDs + flushTable |
| Orphaned rules | ✅ | Same |
| VIP movement (old gw exists) | ✅ | syncVIPs handles |
| VIP movement (old gw removed) | ⚠️ | Same gap as VIP leak |
| Table ID stability | ✅ | emptyDir persistence |
| Atomicity | ✅ (self-healing) | Table IDs: brief disruption on shift. VIPs: no file concern, but leak gap applies for out-of-scope interfaces |
| Safety (no false removals) | ⚠️ | /32 scan heuristic; risk limited to interfaces in current ENC scope |

The PR solves the primary issues (table ID shift, most VIP leaks on in-scope interfaces) and is acceptable for alpha. The identified gaps (removed-gateway VIP cleanup, scan scope safety) are rooted in the fundamental limitation of VIP ownership attribution and can be partially mitigated by interface persistence (proposal A) and interface filter (proposal B) in future iterations.

### Author's Self-Assessment Validation

1. Use struct{} instead of bool: ✅ Valid.
2. OOM crash with stale mapping file: ✅ Valid. Stale gateway cleanup self-heals on first reconcile.
3. VIP operations independent of mapping file: ✅ Valid. Mapping file only affects table IDs, not VIP logic.
4. Mapping file doesn't drive VIP deletion: ✅ Valid. VIP removal driven by managedVIPs (from kernel scan) vs desired (from ENC).
5. VIP moved between gateways: ⚠️ Partially valid. Correct when both gateways still exist in ENC. Gap when old gateway is removed entirely — no syncVIPs called for its exclusive 
interface.

internal/controller/sidecar/persistence.go
```diff
@@ -0,0 +1,81 @@
...
+// interfaces) to enable cleanup of stale VIPs even if the mapping file is lost.
+package sidecar
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+)
+
+const mappingFile = "/var/run/meridio/table-id-mapping.json"
```
zolug: Would it make sense to make this configurable? For example to make disabling the persistence logic explicit and cleaner? Applications might want to use a different emptyDir mount path as well.

IMHO documentation alone could be sufficient for alpha, but configurable is better for production.
Which brings us to a more generic comment: documentation should be updated (e.g. controllers/sidecar.md)

zolug (approved):

Issue #159, opened 2026-05-28
E2E: add dual-stack test suite


zolug:
## Problem

All existing e2e suites use IPv4-only GatewayConfigurations (`internalSubnets` with a single IPv4 CIDR). There is no end-to-end validation that dual-stack Gateways work correctly — both IPv4 and IPv6 VIPs advertised, traffic distributed, and return paths plumbed.

## Proposal

Add a new e2e suite `suites/dual-stack/` with:

- Kind cluster config with dual-stack networking enabled
- NADs supporting both IPv4 and IPv6 on the application network
- GatewayConfiguration with two `internalSubnets` (one per family)
- L34Routes with both IPv4/32 and IPv6/128 VIPs
- GatewayRouters with IPv6 addresses
- Traffic validation for both families (ctraffic already listens on `[::]`)

## Why a separate suite

- Dual-stack requires different infrastructure (Kind config, NADs, external router setup)
- Keeps single-stack and dual-stack failures isolated
- Can be run independently in CI
- Test Go code (`e2e_suite_test.go`) is shared — only YAML manifests differ

## Context

- #144 added unit test coverage for dual-stack endpoint processing in the LB controller
- #128 tracks the broader IP family mode management study (blackhole safety net)
- The DG controller already produces per-family EndpointSlices (#70)

PR #161, opened 2026-05-28
Enhance LB controller for dual-stack IP processing and deduplication

### LINKED ISSUE #144: LB controller: support dual-stack endpoint processing
## Description

The LB controller needs to handle both IPv4 and IPv6 addresses per Maglev identifier to complete dual-stack support. The DG controller is expected to produce consistent Maglev IDs across IPv4 and IPv6 EndpointSlices (#70), but the LB side requires adjustments to consume them correctly.

## Required Changes

- Accumulate addresses per identifier across EndpointSlices (or classify by `AddressType`)
- Create routes for both IP families per fwmark (IPv4 and IPv6 routing tables are independent in Linux)
- Handle per-family route lifecycle (add/update/remove independently)

## Context

- The DG allocation fix (#70) should ensure that the same Pod gets the same Maglev ID in both IPv4 and IPv6 EndpointSlices
- The LB refactor is ideally expected to address this alongside other improvements
- IP family mode management (#128) and blackhole safety nets are related but separate concerns

## Related

- #70 — Maglev ID allocation fix (DG side)
- #83 — LB controller refactor (remove nspAPI dependency)
- #128 — IP family mode management (study)
- #129 — LB routing bug (stale routes)
- Design study: `docs/design/maglev-id-shared-allocation.md` (LB Controller: Explored Changes)

zolug:
Fixes #144

## Changes

Dual-stack endpoint processing:
- Accumulate IPs per identifier across IPv4/IPv6 EndpointSlices (append instead of overwrite)
- Sort EndpointSlices by name for deterministic processing order during multi-slice transients
- Sort IPs before AddTarget to prevent false "IPs changed" triggers from ordering differences

Bug fix — duplicate ip rules:
- ensureRule compared Priority=-1 (vishvananda/netlink sentinel for "unset") against kernel-assigned positive priorities, never finding a match — adding a new rule on every reconcile
- Fix: skip Priority comparison when desired.Priority < 0

## Tests

- Dual-stack AddTarget test (new target + drift recovery path) in internal/nfqlb
- Dual-stack IP accumulation test in LB controller

Verified manually in Kind.

dedicated issue to handle e2e tests: https://github.com/Nordix/Meridio-2/issues/159

### DIFF
--- a/api/v1alpha1/distributiongroup_types.go
+++ b/api/v1alpha1/distributiongroup_types.go
@@ -56,7 +56,7 @@ type MaglevConfig struct {
 	// MaxEndpoints is the table capacity. This field is immutable after creation.
 	//
 	// Immutability rationale:
-	// - Maglev IDs are assigned per DistributionGroup and per IP family (IPv4/IPv6)
+	// - Maglev IDs are assigned per DistributionGroup and per Gateway
 	// - The same Pod IP may have different Maglev IDs in different DistributionGroups
 	// - Changing MaxEndpoints causes Maglev hash table reshuffle, potentially reassigning
 	//   IDs for many endpoints and disrupting active connections for this DistributionGroup
--- a/config/crd/bases/meridio-2.nordix.org_distributiongroups.yaml
+++ b/config/crd/bases/meridio-2.nordix.org_distributiongroups.yaml
@@ -78,7 +78,7 @@ spec:
                       MaxEndpoints is the table capacity. This field is immutable after creation.
 
                       Immutability rationale:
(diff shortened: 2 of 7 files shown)


ljkiraly (approved):
Correct fix for dual-stack IP processing.
Contains the solution for the `slicesEqual` order-sensitivity (sorting before comparison).
Good test coverage.

**Minor issue**: The fix uses `< 0` which is correct for the `-1` sentinel, but the comment could note that `netlink.NewRule()` specifically sets `Priority: -1`. If the library ever changes the sentinel value, this would break. Low risk — the library has used `-1` since inception.

PR #162, opened 2026-05-29
Expose parameters to be set via command line.

arp-est:
https://github.com/Nordix/Meridio-2/issues/116

### DIFF
--- a/docs/controllers/router.md
+++ b/docs/controllers/router.md
@@ -78,7 +78,7 @@ BIRD runs as a child process of the router binary, started via `exec.CommandCont
 - BIRD logs to a configurable file path via `--bird-log-file` flag (default: `/var/log/bird/bird.log`)
 - The config template conditionally emits `log "<path>" all;` when a log file is set
 - `log stderr all;` is always present as baseline
-- The `Bird` struct uses the options pattern: `bird.New(bird.WithLogFile(cfg.BirdLogFile))`
+- The `Bird` struct takes a `bird.Config` struct: `bird.New(bird.Config{...})`
 
 **Interface**: `BirdInterface` allows mocking in tests (config generation, monitoring) without filesystem or process dependencies.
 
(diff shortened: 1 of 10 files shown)


docs/controllers/router.md
```diff
@@ -138,13 +171,7 @@ The monitoring goroutine will evolve from log-only to a data source for the cont
 
 Several values are currently hardcoded and should be exposed via CLI flags / environment variables:
 
-- **Monitoring interval**: Currently `1 * time.Second` in `monitorConnectivity()`
-- **Kernel table IDs**: `defaultKernelTableID` (4096), `blackholeKernelTableID` (4097)
-- **Rule priorities**: `rulePriority` (100), `blackholeRulePriority` (101)
-- **BIRD paths**: Socket path (`/var/run/bird/bird.ctl`), config file (`/etc/bird/bird.conf`)
 - **BGP defaults**: Default ports (`defaultLocalPort`, `defaultRemotePort` = 179), default hold time (90s) — currently constants in `config.go`
```
zolug: Is there a separate task for exposing the BGP related params?

internal/bird/bird.go
```diff
@@ -40,9 +40,21 @@ type BirdInterface interface {
 	Monitor(ctx context.Context, interval time.Duration) (<-chan MonitorStatus, error)
 }
 
+// Default values for configurable parameters.
+const (
```
zolug: This is an internal package. I'm not aware of any motives/plans to transform it to a pkg so that it could be loaded externally. Thus, maintaining "second-line" of hardcoded values seems to be an unnecessary indirection, and means two places to maintain (prone to drift).
Hence, the option pattern is not really jutified either imho. AFAIK for internal packages the preferred approach is plain struct fields or constructor parameters. The cmd layer already knows the values and always passes them.
So a cleaner approach would be `bird.New(config bird.Config)` where Config is a struct with required fields populated by the cmd layer (from the parsed config). (For safety reasons the constructor could validate for missing values and return an error.)
What do you think? Feel free to disagree :)

docs/controllers/router.md
arp-est: I don't think so, I mean you can set the bgp port in the gatewayrouter yaml, so there's not much point in exposing it.

arp-est: want me to remove that line?

internal/bird/bird.go
arp-est: hm, yeah I was also a bit conflicting on having two places for defaults, I think I went with the flow on this one, because there's the LBReadiness, which is used in both the lb, and the router, and this I think already had an options pattern, but I think I like your point, it does make things less convoluted.

docs/controllers/router.md
zolug: I was asking because it seemed that these are similar values but had been ignored. Either they should be exposed and handled similarly as the other values or decided if they are needed at all. Btw, according to GatewayRouter spec the ports are optional, so is the hold time. Meaning they require a default value.

internal/bird/bird.go
```diff
@@ -40,36 +40,32 @@ type BirdInterface interface {
...
-func New(opts ...Option) *Bird {
-	b := &Bird{
-		SocketPath: "/var/run/bird/bird.ctl",
-		ConfigFile: "/etc/bird/bird.conf",
-		nl:         &netlink.Handle{},
-	}
-	for _, o := range opts {
-		o(b)
+func New(cfg Config) (*Bird, error) {
+	if cfg.SocketPath == "" || cfg.ConfigFile == "" || cfg.TableID == 0 || cfg.RulePriority == 0 || cfg.KernelScanTime == 0 {
```
zolug: IIRC rule priority 0 is a valid prio, cannot be used as not set.
As for routing table id, there are reserved tables: 253 is default, 254 is main, 255 is local. If you wish, you check for these as well to reject (while also considering table_id +1 for the blackhole thing).

Otherwise, IMHO looks good.

zolug (approved):

szvincze:
The only thing worth confirming if the HoldTime increase from 90s to 240s was intentional or oversight because 240s is far from the RFC 4271 default of 90s. If it was deliberately changed then I don't have more questions, @zolug has already brought up everything else I found.

arp-est:
> The only thing worth confirming if the HoldTime increase from 90s to 240s was intentional or oversight because 240s is far from the RFC 4271 default of 90s. If it was deliberately changed then I don't have more questions, @zolug has already brought up everything else I found.

With Zoli we talked about that, it should be the same default as the internal guys are using, also bird itself uses this as the default. I'm honestly fine with both of the values as default, its just that you need to choose one.

szvincze (approved):
LGTM

PR #163, opened 2026-05-29
Fix DeleteTarget partial failure: retry on next reconcile

### LINKED ISSUE #151: LB controller; Partial failure handling in multi-step DeleteTarget
### Overview
`AddTarget` and `DeleteTarget` are multi-step operations (nfqlb CLI activation/deactivation + route/rule creation/deletion). Partial failures leave the system in an intermediate state.

### Target Tracking
There are two levels of target tracking that should mirror each other but are updated independently:

1. `c.targets[dgName][identifier]` (Controller level) — used for diffing old vs new in `reconcileTargets`, bulk-committed at end of reconcile 
2. `s.targets[identifier]` (nfqlb Instance level) — used inside `AddTarget`/`DeleteTarget` to determine code path, updated per-operation

`AddTarget` partial failures self-heal in all cases: if activate fails, s.targets is not updated → retry re-attempts activation. If activate succeeds but route creation fails, s.targets is committed → on requeue the idempotent re-apply path fixes routes. If the target disappears from `EndpointSlices` between retries, `c.targets` (committed unconditionally) ensures `DeleteTarget` is called to clean up.

`DeleteTarget` partial failures do not self-heal: `delete(s.targets, identifier)` executes before `doExec("deactivate")`. If deactivation fails:

`s.targets` no longer has the identifier (already deleted) `c.targets[DG] = newTargets` is committed unconditionally (also doesn't have it)
Both maps agree "it's gone" — but nfqlb still has the target active in the hash table
On requeue, the identifier appears in neither `currentTargets` nor `newTargets` → nobody calls `DeleteTarget` again → stuck
The semantics of an identifier being present in s.targets are unclear — does it mean "fully configured" or just "known"? This ambiguity makes it hard to reason about recovery from partial states. Fixing is non-trivial: deferring `c.targets` commit until success doesn't work cleanly because `EndpointSlice` content may change between retries, making the old c.targets stale for diffing.

The probability is low (nfqlb CLI failure for a running process is rare), and the impact is limited (one hash table slot with no route → packet dropped for that bucket). But it's a correctness gap worth acknowledging.

### Possible Remedy
The targets "not fully configured" should be stamped and re-configured at re-conciliation.

ljkiraly:
Two changes to ensure failed deletions are retried:

1. nfqlb: move delete(s.targets, identifier) to after successful deactivation. If deactivation fails, identifier stays in s.targets so AddTarget detects it on retry.

2. LB controller: don't commit c.targets on error. On requeue, the diff against stale currentTargets will call DeleteTarget again for identifiers that failed. Successful adds are idempotent on retry.

Readiness file is updated before the error check so VIP advertisement reflects actual target availability regardless of deletion failures.

Fixes: gh-151

### DIFF
--- a/docs/design/nfqlb-target-recovery-invariants.md
+++ b/docs/design/nfqlb-target-recovery-invariants.md
@@ -1,0 +1,60 @@
+# NFQLB Target Management: Recovery Invariants
+
+## Overview
+
+The LB controller manages NFQLB targets (activate/deactivate slots, create/delete policy routes). The recovery model relies on several implicit invariants that must be preserved for self-healing to work correctly.
+
+## Invariants
+
+### 1. Operation ordering: routes before activate
+
+`AddTarget` creates policy routes before activating the NFQLB slot. This order ensures:
+- If routes fail → no nfqlb slot is activated, no traffic reaches partial routes
+- If activate fails after routes → target is marked broken, routes exist but harmless (invariant #3)
+- On retry, routes are re-applied idempotently then activate is retried
+
+This order is preferred because an active slot with missing routes would cause traffic drops, while routes without an active slot receive no traffic (NFQLB hasn't marked packets with the fwmark yet).
+
+### 2. No context timeouts on nfqlb CLI calls
+
+The `doExec` calls to `nfqlb activate`/`deactivate` use the reconcile context without additional timeouts. This guarantees an unambiguous outcome: the call either succeeds or fails, never times out mid-operation. Adding a custom timeout would break the recovery model — a timed-out activate might have succeeded in NFQLB's shared memory, but the caller wouldn't know whether to retry or clean up.
+
+**Constraint**: Do not wrap nfqlb CLI calls in shortened-timeout contexts.
+
+### 3. Orphaned routes are harmless
+
+Policy routes (fwmark-based) only receive traffic if NFQLB actively sets the corresponding fwmark on packets. A route without an active NFQLB slot never gets traffic. This means:
+- Failed route deletions don't cause traffic blackholes
+- Stale routes from a previous activation are overwritten by `RouteReplace` on retry
+
+**Dependency**: This relies on the fwmark routing design where NFQLB is the sole source of fwmarks. If any other component sets fwmarks in the same range, orphaned routes could receive unintended traffic.
+
+### 4. `s.targets` commit position determines recovery behavior
+
+- `AddTarget`: commits `s.targets[id] = ips` on any partial progress (routes created or activate attempted). If any step fails, the target is in `s.targets` and marked broken. On retry, the "IPs unchanged" path re-applies routes and retries activate if needed.
+- `DeleteTarget`: removes `delete(s.targets, id)` only after all operations succeed (deactivate + route deletion). If any step fails, the identifier stays in `s.targets` and is marked broken so the next reconcile will retry `DeleteTarget`.
+
+The reconciler layer always commits `c.targets = newTargets ∪ failedDeletes`, ensuring the diff on next reconcile triggers DeleteTarget for disappeared-but-failed targets.
+
+### 5. Idempotent nfqlb and route operations
+
+The recovery model relies on:
+- `nfqlb activate` on an already-active slot: no-op (returns success)
+- `nfqlb deactivate` on an inactive slot: no-op (returns success)
+- `RouteReplace`: overwrites existing routes (idempotent by design)
+- `ensureRule`: skips if rule already exists (checks before adding)
+- `doDeletePolicyRoute`: tolerates ENOENT/ESRCH (route/rule not found)
+
+If any of these operations became non-idempotent (e.g., returning errors for "already exists" or "not found"), retries would fail or loop indefinitely.
+
+## Broken State Tracking
+
+The `broken` map (`map[int]struct{}`) tracks identifiers where activate succeeded but route creation failed. This prevents the stuck state where:
+1. AddTarget: activate succeeds, routes fail → target in `s.targets`, marked broken
+2. Target disappears from EndpointSlices before retry
+3. Without broken tracking: nobody calls DeleteTarget (identifier not in stale `c.targets` nor `newTargets`)
+4. With broken tracking: reconcile loop cleans up broken targets not in `newTargets`
+
+The broken flag is cleared on:
+- Successful route creation (all paths: new, same-IPs, changed-IPs)
+- DeleteTarget (regardless of outcome)
--- a/docs/design/nfqlb-target-recovery-invariants.md
+++ b/docs/design/nfqlb-target-recovery-invariants.md
@@ -17,9 +17,7 @@ This order is preferred because an active slot with missing routes would cause t
 
 ### 2. No context timeouts on nfqlb CLI calls
 
-The `doExec` calls to `nfqlb activate`/`deactivate` use the reconcile context without additional timeouts. This guarantees an unambiguous outcome: the call either succeeds or fails, never times out mid-operation. Adding a custom timeout would break the recovery model — a timed-out activate might have succeeded in NFQLB's shared memory, but the caller wouldn't know whether to retry or clean up.
-
-**Constraint**: Do not wrap nfqlb CLI calls in shortened-timeout contexts.
+The `doExec` calls to `nfqlb activate`/`deactivate` use the reconcile context without additional timeouts. This guarantees an unambiguous outcome: the call either succeeds or fails, never times out mid-operation. A timed-out activate might have succeeded in NFQLB's shared memory, but the caller wouldn't know whether to retry or clean up. With broken-state tracking, a timeout would be handled as a broken state (mark broken, retry on next reconcile), so custom timeouts would not break recovery — but they add complexity without clear benefit.
 
 ### 3. Orphaned routes are harmless
 
--- a/docs/design/nfqlb-target-recovery-invariants.md
+++ b/docs/design/nfqlb-target-recovery-invariants.md
@@ -49,12 +49,24 @@ If any of these operations became non-idempotent (e.g., returning errors for "al
 
 ## Broken State Tracking
 
-The `broken` map (`map[int]struct{}`) tracks identifiers where activate succeeded but route creation failed. This prevents the stuck state where:
-1. AddTarget: activate succeeds, routes fail → target in `s.targets`, marked broken
+The `broken` map (`map[int]struct{}`) tracks identifiers in an inconsistent state — where any operation in AddTarget or DeleteTarget partially failed. An identifier is marked broken when:
+
+1. **AddTarget — route creation fails** (activate never attempted): routes may be partially created, no nfqlb slot active. Target committed to `s.targets` as broken.
+2. **AddTarget — activate fails** (routes were created successfully): routes exist, nfqlb slot state unknown. Target committed to `s.targets` as broken.
+3. **AddTarget — IPs-unchanged or IPs-changed re-apply path**: route or activate retry fails. Marked broken.
+4. **DeleteTarget — deactivate fails**: slot may still be active, routes still exist. Stays in `s.targets`, marked broken.
+5. **DeleteTarget — route deletion fails** (deactivate succeeded): slot inactive, routes partially remain. Stays in `s.targets`, marked broken.
+
+Without broken tracking, the following stuck state occurs:
+1. AddTarget: routes fail → target in `s.targets`, no retry trigger
 2. Target disappears from EndpointSlices before retry
-3. Without broken tracking: nobody calls DeleteTarget (identifier not in stale `c.targets` nor `newTargets`)
+3. Nobody calls DeleteTarget (identifier not in stale `c.targets` nor `newTargets`)
 4. With broken tracking: reconcile loop cleans up broken targets not in `newTargets`
 
-The broken flag is cleared on:
-- Successful route creation (all paths: new, same-IPs, changed-IPs)
-- DeleteTarget (regardless of outcome)
+The broken flag is cleared only when **all** operations in AddTarget or DeleteTarget complete successfully:
+- AddTarget: routes created + activate succeeded (applies to all paths: new, same-IPs, IPs-changed)
+- DeleteTarget: deactivate + all route deletions succeeded → entry removed from both maps
+
+All paths (same-IPs, IPs-changed) include an `activate` call after route creation. Since `activate` is idempotent (no-op on already-active slot per invariant #5), this ensures recovery when the previous failure was an activate failure — not just a route failure.
+
+**Note on DeleteTarget clearance**: The `delete(s.broken, identifier)` at the end of a successful DeleteTarget also removes the target from `s.targets`. If DeleteTarget fails at any step, the identifier remains in both `s.targets` and `broken`, ensuring the next reconcile retries the deletion.
--- a/internal/nfqlb/nfqlb.go
+++ b/internal/nfqlb/nfqlb.go
@@ -223,7 +223,7 @@ type Instance struct {
 	*nfqlbInstanceConfig
 	name                              string
 	targets                           map[int][]string // Key: identifier ; Value: IPs
-	broken                            map[int]struct{} // identifiers activated but with incomplete routes
+	broken                            map[int]struct{} // identifiers in inconsistent state (partial route or activate failure)
 	offset                            int
 	mu                                sync.Mutex
 	updateNfQueueDestinationCIDRsFunc func(ctx context.Context) error
@@ -477,8 +477,9 @@ func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int) 
 	existingIPs, exists := s.targets[identifier]
 	if exists {
 		if slicesEqual(existingIPs, ips) {
-			// IPs unchanged — re-apply routes to recover from potential kernel drift.
-			// RouteReplace and ensureRule are idempotent — no traffic disruption.
+			// IPs unchanged — re-apply routes and re-activate to recover from
+			// potential kernel drift or a previously broken state.
+			// RouteReplace, ensureRule, and activate are all idempotent.
 			fwmark := identifier + s.offset
 			var errFinal error
 			for _, ip := range ips {
@@ -488,10 +489,19 @@ func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int) 
 			}
 			if errFinal != nil {
 				s.broken[identifier] = struct{}{}
-			} else {
-				delete(s.broken, identifier)
+				return errFinal
 			}
-			return errFinal
+			output, err := s.doExec(ctx, "activate",
+				fmt.Sprintf("--index=%d", identifier),
+				fmt.Sprintf("--shm=%s", s.name),
+				strconv.Itoa(identifier+s.offset),
+			)
+			if err != nil {
+				s.broken[identifier] = struct{}{}
+				return fmt.Errorf("failed activating nfqlb target ; %w; %s", err, output)
+			}
+			delete(s.broken, identifier)
+			return nil
 		}
 		// IPs changed — clean old neighbors, delete old routes, apply new ones
 		ctrl.LoggerFrom(ctx).Info("nfqlb: target IPs changed, updating routes",
@@ -516,15 +526,25 @@ func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int) 
 		}
 		if errFinal != nil {
 			s.broken[identifier] = struct{}{}
-		} else {
-			delete(s.broken, identifier)
+			return errFinal
 		}
-		return errFinal
+		output, err := s.doExec(ctx, "activate",
+			fmt.Sprintf("--index=%d", identifier),
+			fmt.Sprintf("--shm=%s", s.name),
+			strconv.Itoa(identifier+s.offset),
+		)
+		if err != nil {
+			s.broken[identifier] = struct{}{}
+			return fmt.Errorf("failed activating nfqlb target ; %w; %s", err, output)
+		}
+		delete(s.broken, identifier)
+		return nil
 	}
 
 	ctrl.LoggerFrom(ctx).Info("nfqlb: add target", "instance", s.name, "ips", ips, "identifier", identifier)
 
 	fwmark := identifier + s.offset
+	s.targets[identifier] = ips
 
 	// Create policy routes first — if this fails, no nfqlb slot is activated.
 	var errFinal error
@@ -534,7 +554,6 @@ func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int) 
 		}
 	}
 	if errFinal != nil {
-		s.targets[identifier] = ips
 		s.broken[identifier] = struct{}{}
 		return errFinal
 	}
@@ -545,12 +564,10 @@ func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int) 
 		strconv.Itoa(identifier+s.offset),
 	)
 	if err != nil {
-		s.targets[identifier] = ips
 		s.broken[identifier] = struct{}{}
 		return fmt.Errorf("failed activating nfqlb target ; %w; %s", err, output)
 	}
 
-	s.targets[identifier] = ips
 	delete(s.broken, identifier)
 	ctrl.LoggerFrom(ctx).Info("nfqlb: target added", "instance", s.name, "ips", ips, "identifier", identifier)
 
--- a/internal/nfqlb/nfqlb.go
+++ b/internal/nfqlb/nfqlb.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"maps"
 	"net"
 	"os/exec"
 	"strconv"
@@ -222,6 +223,7 @@ type Instance struct {
 	*nfqlbInstanceConfig
 	name                              string
 	targets                           map[int][]string // Key: identifier ; Value: IPs
+	broken                            map[int]struct{} // identifiers activated but with incomplete routes
 	offset                            int
 	mu                                sync.Mutex
 	updateNfQueueDestinationCIDRsFunc func(ctx context.Context) error
@@ -271,6 +273,7 @@ func (nfqlb *NFQueueLoadBalancer) AddInstance(ctx context.Context,
 		name:                              name,
 		nfqlbInstanceConfig:               config,
 		targets:                           map[int][]string{},
+		broken:                            map[int]struct{}{},
 		updateNfQueueDestinationCIDRsFunc: nfqlb.updateNfQueueDestinationCIDRs,
 		offset:                            offset,
 		nfqlbPath:                         nfqlb.nfqlbPath,
@@ -483,6 +486,11 @@ func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int) 
 					errFinal = errors.Join(errFinal, err)
 				}
 			}
+			if errFinal != nil {
+				s.broken[identifier] = struct{}{}
+			} else {
+				delete(s.broken, identifier)
+			}
 			return errFinal
 		}
 		// IPs changed — clean old neighbors, delete old routes, apply new ones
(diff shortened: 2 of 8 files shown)


ljkiraly:
Adopting the sidecar pattern adds complexity for marginal benefit. The current "retry everything" approach is 
simpler and the idempotency guarantees make it safe.

zolug:
Introduces regression in the following scenario (probably not a major issue considering how likely it is to happen):
Add fails (activate succeeds, route creation fails), target disappears from EndpointSlices on requeue. -> s.targets HAS entry. c.targets NOT committed (stale, no identifier). On requeue, identifier in neither currentTargets (stale) nor newTargets (target gone). Deactivate loop doesn't trigger. Activate loop doesn't trigger. Nobody calls DeleteTarget. s.targets keeps the entry, nfqlb remains active, partial routes persist. Stuck.


Otherwise in general the current design relies on implicit invariants that aren't documented or enforced:

 1. **Operation ordering matters** — activate-before-routes vs routes-before-activate could have different failure/recovery properties, but nothing in the code or comments explains why the chosen order is correct
 2. **No custom context timeouts** — the "no ambiguous outcome" guarantee of nfqlb activation/deactivation is an assumption, not enforced. Anyone adding a timeout breaks the recovery model silently.
3. **Orphaned routes are harmless** — true only because the current fwmark-based policy routing design ensures traffic only reaches a route if nfqlb actively sets the corresponding fwmark. This is a property of the routing setup, not a universal guarantee — not documented as a dependency
4. **`s.targets` commit position is critical** — the exact placement of `delete(s.targets, id)` or `s.targets[id] = ips` relative to the nfqlb CLI call determines whether the system can self-heal on failure. Moving it one line changes recovery behavior entirely
5. **Idempotent nfqlb and route operations** — the recovery model relies on nfqlb deactivate being a no-op on an inactive slot, RouteReplace overwriting existing routes, and ensureRule skipping existing rules. If any of these operations became non-idempotent (e.g., returning errors for "already exists" or "not found"), retries would fail or loop indefinitely.

IMHO these should be documented to avoid problems (i.e. breaking the code) in the future, or at least to remember :)

Also, currently deletePolicyRoute returns error on "not found" and callers use _ = to ignore ALL errors — including real ones (permission denied, invalid argument). This seems to be wrong:
1. It's our internal API, we define the contract. "Not found" errors shouldn't be errors at all (desired state already achieved)
2. Real errors are silently swallowed


Created a comparision for the use-cases I identified:
```
Three options compared:

- **A: Master (pre PR #163)** — activate-before-routes in AddTarget, delete(s.targets) before deactivate in DeleteTarget, c.targets committed unconditionally
- **B: PR #163** — activate-before-routes in AddTarget, delete(s.targets) after deactivate in DeleteTarget, c.targets committed only on success
- **C: PR #163 + routes-before-activate** — routes-before-activate in AddTarget, delete(s.targets) after deactivate in DeleteTarget, c.targets committed only on success

Use case 1: Add succeeds fully

A: ✅  
B: ✅  
C: ✅

Use case 2: Add fails, target still desired on requeue, same IPs

Route fails:  
A: s.targets committed (activate succeeded first), routes partial. Requeue → AddTarget → s.targets has id, same IPs → re-apply routes. ✅ But blackhole window exists (nfqlb active, 
routes missing until retry).  
B: Same as A. ✅ Same blackhole window.  
C: s.targets empty (routes fail before activate). Requeue → new-target path → RouteReplace + activate. ✅ No blackhole window.

Activate fails:  
A: s.targets empty (activate failed before commit). No routes (routes come after activate). Requeue → new-target path → retries. ✅  
B: Same as A. ✅  
C: s.targets empty. Routes exist (created before activate). Requeue → new-target path → RouteReplace (idempotent) → activate retries. ✅

Use case 3: Add fails, target disappears on requeue

Route fails:  
A: s.targets committed (activate succeeded). c.targets committed with id. Target gone. Requeue: id in currentTargets, not in newTargets → DeleteTarget → s.targets has it → deactivates 
+ deletes routes. ✅  
B: s.targets committed. c.targets NOT committed (stale, no id). Target gone. Requeue: id in neither map. Nobody cleans. nfqlb active, routes partial. Stuck. ⚠️  
C: s.targets empty (routes fail before activate). nfqlb inactive. c.targets not committed. Target gone. Requeue: id in neither map. Partial routes orphaned but nfqlb inactive → no 
traffic. Harmless leak. ✅

Activate fails:  
A: s.targets empty. No routes. nfqlb inactive. c.targets committed with id. Target gone. Requeue: id in currentTargets, not in newTargets → DeleteTarget → s.targets empty → returns 
nil. Nothing to clean (nothing was created). ✅  
B: s.targets empty. No routes. nfqlb inactive. c.targets not committed. Target gone. Requeue: id in neither map. Nothing active. ✅  
C: s.targets empty. Routes exist. nfqlb inactive. c.targets not committed. Target gone. Requeue: id in neither map. Routes orphaned, nfqlb inactive. Harmless leak. ✅

Use case 4: Add fails, target still desired on requeue, IPs changed (same families)

Route fails:  
A: s.targets committed with old IPs (activate succeeded). Requeue with new IPs → AddTarget → s.targets has id, different IPs → "IPs changed" path → deletes old routes, creates new. ✅ 
(blackhole window until retry though)  
B: Same as A. ✅ Same blackhole window.  
C: s.targets empty. Requeue → new-target path → RouteReplace overwrites per-family. ✅ No blackhole.

Activate fails:  
A: s.targets empty. No routes. Requeue → new-target path with new IPs. ✅  
B: Same. ✅  
C: s.targets empty. Routes for old IPs exist. Requeue with new IPs → new-target path → RouteReplace overwrites (same table, same family). ✅

Use case 4 variant: IP family removed between attempts

Route fails:  
A: s.targets committed with old IPs (dual-stack). Requeue with IPv4 only → AddTarget → different IPs → "IPs changed" path → deletes old IPv6 routes, creates IPv4 only. ✅  
B: Same as A. ✅  
C: s.targets empty. First attempt partially created routes (e.g., IPv6 succeeded, IPv4 failed). Requeue with IPv4 only → new-target path → creates IPv4 (RouteReplace). Does NOT delete 
stale IPv6 rule+route (no "old IPs" known). After activation, IPv6 traffic misrouted via stale route. Problem. ⚠️

Activate fails:  
A: s.targets empty. No routes (activate-first, failed before routes). Requeue with fewer families → new-target path → only creates needed family. ✅  
B: Same. ✅  
C: s.targets empty. Routes for both families exist (step 1 succeeded). Requeue with IPv4 only → new-target path → creates IPv4 (RouteReplace). Stale IPv6 rule+route persist. After 
activation, IPv6 misrouted. Problem. ⚠️

Use case 4 variant: IP family added between attempts

All three options: s.targets empty (or has old single-family IPs for A/B) → either "IPs changed" path or new-target path creates routes for both families. ✅ for all.

Use case 5: Delete succeeds fully

A: ✅  
B: ✅  
C: ✅

Use case 6: Delete fails, target still absent on requeue

Deactivation fails:  
A: delete(s.targets) executed before deactivate. c.targets committed without id. Requeue: id in neither map. Nobody retries. Stuck. ❌  
B: s.targets kept. c.targets not committed (stale, has id). Requeue: id in currentTargets, not in newTargets → retries. ✅  
C: Same as B. ✅

Deactivation succeeds, route deletion fails:  
A: delete(s.targets) executed. Routes use _ =. Returns nil. c.targets committed. Stale routes, nfqlb inactive. Minor leak. ✅  
B: Same. ✅  
C: Same. ✅

Use case 7: Delete fails, target reappears on requeue, same IPs

Deactivation fails:  
A: delete(s.targets) executed. c.targets committed without id. Requeue: id not in currentTargets, IS in newTargets → AddTarget → s.targets empty → new-target path → activates (
overwrites still-active slot). ✅  
B: s.targets kept. c.targets not committed (has id). Requeue: id in both maps → AddTarget → s.targets has it, same IPs → re-apply routes. nfqlb still active. ✅  
C: Same as B. ✅

Deactivation succeeds, route deletion fails:  
A: delete(s.targets). Returns nil. c.targets committed without id. Requeue → AddTarget → new-target path → routes + activate. ✅  
B: Same. ✅  
C: Same (routes-before-activate for new-target path, then activate). ✅

Use case 8: Delete fails, target reappears on requeue, different IPs

Deactivation fails:  
A: delete(s.targets). c.targets committed without id. Requeue with new IPs: id not in currentTargets, IS in newTargets → AddTarget → s.targets empty → new-target path → activate (
overwrites) + routes for new IPs. Old routes for old IPs? With A (activate-first): old routes were created after activate on original Add, so they exist. New AddTarget (activate-first)
→ activate overwrites → then routes for new IPs (RouteReplace overwrites same-family). If family removed, stale route from original Add persists. Minor concern if families changed. ⚠️  
B: s.targets kept with old IPs. c.targets not committed. Requeue with new IPs: id in both maps → AddTarget → s.targets has it, different IPs → "IPs changed" path → deletes old routes, 
creates new. ✅  
C: Same as B. ✅

Deactivation succeeds, route deletion fails:  
A: delete(s.targets). Returns nil. c.targets committed. Requeue with new IPs: new-target path → routes (RouteReplace overwrites same-family stale routes) → activate. If family removed,
stale route persists. Minor concern. ⚠️  
B: Same as A (deactivation succeeded → delete(s.targets) → returns nil → c.targets committed). ✅ with same minor concern.  
C: Same, but routes-before-activate. Same minor concern. ⚠️

Final summary:

| Use case | A (master) | B (PR #163) | C (PR #163 + routes-first) |
|----------|-----------|-------------|---------------------------|
| 1. Add succeeds | ✅ | ✅ | ✅ |
| 2. Add fails, target stays, same IPs | ✅ blackhole window | ✅ blackhole window | ✅ no blackhole |
| 3. Add/route fails, target gone | ✅ cleans up | ⚠️ stuck (nfqlb active) | ✅ harmless leak |
| 3. Add/activate fails, target gone | ✅ | ✅ | ✅ harmless leak |
| 4. Add fails, target stays, IPs changed (same families) | ✅ blackhole window | ✅ blackhole window | ✅ no blackhole |
| 4. Add fails, target stays, family removed | ✅ | ✅ | ⚠️ stale route for removed family |
| 4. Add fails, target stays, family added | ✅ | ✅ | ✅ |
| 5. Delete succeeds | ✅ | ✅ | ✅ |
| 6. Delete/deactivate fails, target gone | ❌ stuck | ✅ retries | ✅ retries |
| 6. Delete/route fails, target gone | ✅ minor leak | ✅ minor leak | ✅ minor leak |
| 7. Delete fails, target reappears, same IPs | ✅ | ✅ | ✅ |
| 8. Delete fails, target reappears, different IPs | ⚠️ if families changed | ✅ | ✅ |

Key trade-offs:

- A (master): stuck on delete failure (#151). Clean on add failures (because c.targets committed enables diff-based cleanup).
- B (PR #163): fixes delete stuck-state. Introduces regression when add/route fails AND target disappears (stuck nfqlb activation).
- C (PR #163 + routes-first): fixes delete stuck-state. Eliminates blackhole window. Eliminates B's regression. Introduces stale-route issue when IP family is removed between a failed 
attempt and retry (rare, requires partial route success + family composition change).
```

An alternate idea relying on **broken-state tracking**, where the reconciler layer would not do any optimization mostly to compensate against potential kernel drifts (and to support future neighbor flush optimizations). Reconciler declares intent ("these targets with these IPs"), nfqlb layer handles all operational details (routing, activation, neighbor management, drift recovery, MAC-aware flush decisions etc.) It aims to guarantee eventual convergence (via retry), which macthes the Kubernetes controller pattern as well — reconcile toward desired state, retry on failure. LB process restart is a non-concern: nfqlb shared memory is gone, in-memory state is lost, and startup cleanup removes stale kernel routes/rules — the first reconcile rebuilds everything from scratch.

```
Option D: Broken-state tracking (optimization on nfqlb.go layer)

Key design:
- Reconciler is dumb: always calls AddTarget for all desired ids, DeleteTarget for all unwanted ids. No optimization, no broken flag on reconciler level.
- c.targets = all ids from newTargets (regardless of Add success/failure) ∪ ids from currentTargets whose DeleteTarget failed. This ensures the diff on next reconcile triggers 
DeleteTarget for disappeared targets (even if never successfully added) and retries failed deletions.
- All optimization lives in nfqlb.go layer via s.targets which stores {ips, broken} per identifier.
- No broken flag on c.targets because the reconciler performs no optimization — it always calls Add/Delete unconditionally.
- Error return triggers requeue.

nfqlb.go AddTarget(id, ips):
- No entry → create routes → activate → store {ips, not broken}; on error → store {ips, broken}, return error
- Entry, not broken, same IPs → re-apply routes (drift recovery). Return success.
- Entry, not broken, different IPs → delete old routes → create new routes; on error → mark broken
- Entry, broken → delete routes for stored IPs → create routes for new IPs → activate; on success → {ips, not broken}; on error → {ips, broken}, return error

nfqlb.go DeleteTarget(id):
- No entry → return nil
- Entry → deactivate + delete routes for stored IPs (idempotent, ignore not-found); on success → remove entry; on error → mark broken, return error

Note on sub-case splitting: The broken flag + idempotent operations + full redo from scratch means the recovery path is identical regardless of which step failed. Sub-cases (route 
fails vs activate fails) are not enumerated separately — the broken flag erases the distinction.

Use case 1: Add succeeds

AddTarget: routes → activate → s.targets = {ips, not broken}. Reconciler: no error, c.targets committed (has id from newTargets). ✅

Use case 2: Add fails, target still desired, same IPs

AddTarget fails (any step): s.targets = {ips, broken}. Reconciler: error returned, c.targets committed (has id from newTargets), requeue triggered.

Requeue: reconciler rebuilds newTargets (same id, same IPs), calls AddTarget. nfqlb sees broken → full redo: delete stored IPs routes → create routes → activate → not broken. ✅

Use case 3: Add fails, target disappears

AddTarget fails: s.targets = {ips, broken}. Reconciler: error returned, c.targets committed (has id from newTargets), requeue triggered.

Requeue: newTargets doesn't have id. Id in currentTargets (committed), not in newTargets → reconciler calls DeleteTarget. nfqlb: entry exists (broken) → deactivate (idempotent, no-op 
if never activated) + delete routes (idempotent) → entry removed. ✅

Use case 4: Add fails, target still desired, IPs changed (same families)

AddTarget fails: s.targets = {oldIPs, broken}. Reconciler: c.targets committed (has id), requeue.

Requeue with new IPs: reconciler calls AddTarget(newIPs). nfqlb: broken → delete routes for stored oldIPs → create routes for newIPs → activate → {newIPs, not broken}. ✅

Use case 4 variant: IP family removed

s.targets = {[ipv4, ipv6], broken}. Requeue with [ipv4] only.

AddTarget([ipv4]) → broken → delete routes for stored [ipv4, ipv6] (both families explicitly cleaned) → create [ipv4] route → activate → {[ipv4], not broken}. ✅

Use case 4 variant: IP family added

s.targets = {[ipv4], broken}. Requeue with [ipv4, ipv6].

AddTarget([ipv4, ipv6]) → broken → delete old [ipv4] routes → create both → activate → {[ipv4, ipv6], not broken}. ✅

Use case 5: Delete succeeds

DeleteTarget: deactivate → delete routes → entry removed. Reconciler: no error, c.targets committed (id not present — not in newTargets, delete succeeded so not in failed-deletes). ✅

Use case 6: Delete fails, target still absent

DeleteTarget fails: s.targets kept (broken). Reconciler: error returned, c.targets = newTargets ∪ failed-delete id (preserved), requeue triggered.

Requeue: newTargets doesn't have id. Id in currentTargets (preserved from merge), not in newTargets → DeleteTarget retries. nfqlb: entry exists (broken) → deactivate (idempotent) + 
delete routes (idempotent) → entry removed. ✅

Use case 7: Delete fails, target reappears, same IPs

DeleteTarget fails: s.targets = {ips, broken}. c.targets has id (merged). Requeue: id in both currentTargets and newTargets → not in delete loop → reconciler calls AddTarget(ips). 
nfqlb: broken → full redo: delete routes + create routes + activate → not broken. ✅

Use case 8: Delete fails, target reappears, different IPs

DeleteTarget fails: s.targets = {oldIPs, broken}. c.targets has id (merged). Requeue with new IPs: reconciler calls AddTarget(newIPs). nfqlb: broken → delete routes for stored oldIPs →
create routes for newIPs → activate → {newIPs, not broken}. ✅

All use cases pass. No stuck states, no stale routes, no blackhole windows.

Dependencies:
- nfqlb activate/deactivate are idempotent (verified from nfqlb source)
- RouteReplace/ensureRule are idempotent
- deletePolicyRoute must ignore not-found errors (currently returns error on ENOENT — needs fixing)
- s.targets preserves stored IPs on failure (broken flag retains them for cleanup on retry)
- c.targets = newTargets ∪ failed-delete ids (ensures reconciler retries deletions and cleans up disappeared broken targets)
- Routes-before-activate order in redo path (eliminates blackhole window)

```

internal/nfqlb/nfqlb.go
```diff
@@ -506,34 +514,47 @@ func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int)
...
+		} else {
+			delete(s.broken, identifier)
+		}
 		return errFinal
 	}
 
 	ctrl.LoggerFrom(ctx).Info("nfqlb: add target", "instance", s.name, "ips", ips, "identifier", identifier)
 
+	fwmark := identifier + s.offset
+
```
zolug, line 506: Why not do `s.targets[identifier] = ips` here once?

internal/nfqlb/nfqlb.go
```diff
@@ -483,6 +486,11 @@ func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int)
 					errFinal = errors.Join(errFinal, err)
 				}
 			}
+			if errFinal != nil {
```
zolug: `exists` doesn't necessarily mean that it got activated. It's semantics have changed. If state was broken, activation must be re-attempted.

internal/nfqlb/nfqlb.go
```diff
@@ -222,6 +223,7 @@ type Instance struct {
 	*nfqlbInstanceConfig
 	name                              string
 	targets                           map[int][]string // Key: identifier ; Value: IPs
+	broken                            map[int]struct{} // identifiers activated but with incomplete routes
```
zolug: misleading description

internal/nfqlb/nfqlb.go
```diff
@@ -506,34 +514,47 @@ func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int)
 				errFinal = errors.Join(errFinal, err)
 			}
 		}
+		if errFinal != nil {
```
zolug: Same as before:
`exists` doesn't necessarily mean that it got activated. It's semantics have changed. If state was broken, activation must be re-attempted.

internal/nfqlb/nfqlb.go
ljkiraly: Fixed by commit 5ed39878

internal/nfqlb/nfqlb.go
ljkiraly: Fixed by commit 5ed3987

internal/nfqlb/nfqlb.go
ljkiraly: Fixed by commit 5ed3987

docs/design/nfqlb-target-recovery-invariants.md
```diff
@@ -0,0 +1,72 @@
...
+`AddTarget` creates policy routes before activating the NFQLB slot. This order ensures:
+- If routes fail → no nfqlb slot is activated, no traffic reaches partial routes
+- If activate fails after routes → target is marked broken, routes exist but harmless (invariant #3)
+- On retry, routes are re-applied idempotently then activate is retried
+
+This order is preferred because an active slot with missing routes would cause traffic drops, while routes without an active slot receive no traffic (NFQLB hasn't marked packets with the fwmark yet).
+
+### 2. No context timeouts on nfqlb CLI calls
+
+The `doExec` calls to `nfqlb activate`/`deactivate` use the reconcile context without additional timeouts. This guarantees an unambiguous outcome: the call either succeeds or fails, never times out mid-operation. Adding a custom timeout would break the recovery model — a timed-out activate might have succeeded in NFQLB's shared memory, but the caller wouldn't know whether to retry or clean up.
```
zolug: With the proposed "broken state" based approach a custom context timeout would not break the recovery, however timeout would need to be handled as broken state.
Therefore, no need for the strict constraint "Do not wrap nfqlb CLI calls in shortened-timeout contexts".

internal/nfqlb/nfqlb.go
```diff
@@ -474,16 +477,31 @@ func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int)
 	existingIPs, exists := s.targets[identifier]
 	if exists {
 		if slicesEqual(existingIPs, ips) {
-			// IPs unchanged — re-apply routes to recover from potential kernel drift.
```
zolug, line 477: A lot of duplicated logic in this block now, that could be avoided:
- The "exists" paths only need to handle cleanup of stale state (delete routes for removed IPs), after which they can fall through to the common path that does create routes for current IPs, manages activation and broken flag.
- Also a deferred function based broken flag management would be cleaner and more idiomatic.
- Moreover, if the identifier exists and is not broken, the nfqlb activation already happened successfully before (the only way to reach not-broken state). Since activate is idempotent (same fwmark → no-op in nfqlb), skipping it just saves one CLI call. The routes still need re-applying (drift recovery, IP changes), but the activate call is pure overhead in this case.


So, sg similar:
```go
// Note: With a named return (err error), any return <value> assigns <value> to err before deferred functions run. The defer then reads the final err.
func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int) (err error) {
    // ... validation ...
    s.mu.Lock()
    defer s.mu.Unlock()

    defer func() {
        if err != nil {
            s.broken[identifier] = struct{}{}
        } else {
            delete(s.broken, identifier)
        }
    }()

    existingIPs, exists := s.targets[identifier]
    skipActivate := exists && !s.isBroken(identifier)

  // Handle existing target: clean up stale state (can add logs for failed actions)
  if exists && !slicesEqual(existingIPs, ips) {
        fwmark := identifier + s.offset
        for _, ip := range existingIPs {
            // If any IP changed for an identifier, it might imply a Pod replacement (new Pod, new MAC for all IPs, even those that happen to be the same address)
            // Note: tracking Pod UID per identifier would make this precise rather than heuristic-based
            _ = cleanNeighbor(net.ParseIP(ip))
            // Clean up routes for IPs no longer present
            if !sliceContains(ips, ip) {
                _ = s.doDeletePolicyRoute(fwmark, ip)
            }
        }
    }

    // Update stored IPs and create routes
    s.targets[identifier] = ips
    fwmark := identifier + s.offset

    for _, ip := range ips {
        if e := s.doCreatePolicyRoute(fwmark, ip); e != nil {
            err = errors.Join(err, e)
        }
    }
    if err != nil {
        return
    }

    // Activate only for new targets or recovery from broken state
    if skipActivate {
        return
    }

    var output []byte
    output, err = s.doExec(ctx, "activate",
        fmt.Sprintf("--index=%d", identifier),
        fmt.Sprintf("--shm=%s", s.name),
        strconv.Itoa(identifier+s.offset),
    )
    if err != nil {
        err = fmt.Errorf("failed activating: %w; %s", err, output)
    }

    return
}
```

Or without the activation optimization:
```go
func (s *Instance) AddTarget(ctx context.Context, ips []string, identifier int) (err error) {
    // ... validation ...
    s.mu.Lock()
    defer s.mu.Unlock()

    defer func() {
        if err != nil {
            s.broken[identifier] = struct{}{}
        } else {
            delete(s.broken, identifier)
        }
    }()

    // Handle existing target: clean up stale state (can add logs for failed actions)
    if exists && !slicesEqual(existingIPs, ips) {
        fwmark := identifier + s.offset
        for _, ip := range existingIPs {
            // Note: tracking Pod UID per identifier would make this precise rather than heuristic-based
            _ = cleanNeighbor(net.ParseIP(ip))
             // Clean up routes for IPs no longer present
            if !sliceContains(ips, ip) {
                _ = s.doDeletePolicyRoute(fwmark, ip)
            }
        }
    }

    // Update stored IPs, create routes, activate
    s.targets[identifier] = ips
    fwmark := identifier + s.offset

    for _, ip := range ips {
        if e := s.doCreatePolicyRoute(fwmark, ip); e != nil {
            err = errors.Join(err, e)
        }
    }
    if err != nil {
        return
    }

    var output []byte
    output, err = s.doExec(ctx, "activate",
        fmt.Sprintf("--index=%d", identifier),
        fmt.Sprintf("--shm=%s", s.name),
        strconv.Itoa(identifier+s.offset),
    )
    if err != nil {
        err = fmt.Errorf("failed activating: %w; %s", err, output)
        return
    }

    return
}
```

internal/nfqlb/nfqlb.go
```diff
@@ -574,8 +612,29 @@ func slicesEqual(a, b []string) bool {
 	return true
 }
 
-// DeleteTarget deletes a target identifier to the nfqlb instance
-// and deletes the policy route associated.
+// BrokenTargets returns identifiers that were activated but have incomplete routes.
+// The caller should call DeleteTarget for any broken identifier no longer in the desired set.
+func (s *Instance) BrokenTargets() map[int]struct{} {
```
zolug: Are BrokenTargets() and Targets() methods needed at all?

internal/nfqlb/nfqlb.go
zolug: Deferred function based broken flag management would look cleaner imho:

E.g.:

```go
//With a named return (err error), any return <value> assigns <value> to err before deferred functions run. The defer then reads the final err.
func (s *Instance) deleteTargetNoLock(ctx context.Context, identifier int) (err error) {
    storedIPs, exists := s.targets[identifier]
    if !exists {
        return nil  // err = nil, defer sees nil → success path
    }

    defer func() {
        if err != nil {
            s.broken[identifier] = struct{}{}
        } else {
            delete(s.targets, identifier)
            delete(s.broken, identifier)
        }
    }()

    var output []byte
    output, err = s.doExec(ctx, "deactivate",
        fmt.Sprintf("--index=%d", identifier),
        fmt.Sprintf("--shm=%s", s.name),
    )
    if err != nil {
        err = fmt.Errorf("failed deactivating: %w; %s", err, output)
        return  // defer sees err != nil → marks broken
    }

    for _, ip := range storedIPs {
        if e := s.doDeletePolicyRoute(identifier+s.offset, ip); e != nil {
            err = errors.Join(err, e)
        }
    }

    return  // defer sees err (nil or joined errors)
}

```

docs/design/nfqlb-target-recovery-invariants.md
ljkiraly: Solved in 3c2b2e3

internal/nfqlb/nfqlb.go
ljkiraly: Solved in 3c2b2e3

internal/nfqlb/nfqlb.go
ljkiraly: Solved in 3c2b2e3

internal/nfqlb/nfqlb.go
ljkiraly: Solved in 3c2b2e3

ljkiraly:
Current state matrix created by AI shows no stuck states, no missed cleanups. The code covers:

| Scenario | Route outcome | Activate/Deactivate outcome | Handled? |
|---|---|---|---|
| Add new | ✅ success | ✅ success | ✅ |
| Add new | ❌ fail | — (not attempted) | ✅ broken, retry |
| Add new | ✅ success | ❌ fail | ✅ broken, retry |
| Add existing (drift) | ✅ success | ⏭ skipped | ✅ |
| Add existing (drift) | ❌ fail | — | ✅ broken, full redo on retry |
| Add existing (broken) | ✅ success | ✅ success | ✅ recovered |
| Add existing (broken) | ❌ fail | — | ✅ stays broken |
| Add existing (broken) | ✅ success | ❌ fail | ✅ stays broken |
| Delete | — | ✅ deactivate OK | ✅ routes cleaned |
| Delete | routes ❌ | ✅ deactivate OK | ✅ broken, retry |
| Delete | — | ❌ deactivate fail | ✅ broken, retry |

zolug (approved):

Issue #118, opened 2026-04-29
Make type=Maglev DG's maxTarget field mandatory (no default value)


zolug:
Avoid API Server based default value setting.

### Problem

MaxEndpoints has its default value defined in two places:
- CRD: +kubebuilder:default=32 on the field
- Controller: DefaultMaglevMaxEndpoints int32 = 32 (Go constant)

These can drift silently. It's unclear which is the source of truth.

### Context

The MaglevConfig block is an optional pointer (*MaglevConfig) by design — this supports future extensibility. When new DG types are added (e.g., RoundRobin), the API shouldn't inject empty/default Maglev blocks into non-Maglev DGs. The existing CEL (self.spec.type == 'Maglev' ? true : !has(self.spec.maglev)) prevents the Maglev block from being set on non-Maglev types.

### Tension

- **User-friendly:** Minimal config should work out of the box. A DG with just a selector gets Maglev with 32 endpoints — no need to understand internals.
- **Future-proof:** The pattern must scale to additional DG types without each needing its own defaulting story or polluting other types' config.
- **Single source of truth:** Defaults should live in one place to prevent drift.

### Proposed Direction

Remove +kubebuilder:default=32 from the CRD. Make MaxEndpoints required (non-optional) within MaglevConfig:

- **User omits maglev: block entirely** → controller applies Go constant (32). Minimal config still works.
- **User provides maglev: block** → must explicitly set maxEndpoints. No hidden CRD default. The user made a conscious choice to configure Maglev, so they should state the capacity.

This pattern ("optional type-specific block with required fields within it, controller defaults for absent block") scales to any number of future types. Each type gets its own optional pointer block with required fields. The controller has one constant per type for the "block omitted" case. No CRD defaults within blocks, no webhooks, no cross-type pollution.

### Trade-off

If the Go constant changes in a future controller version, existing DGs that omitted the block silently get different behavior. This is inherent to controller-side defaulting and mitigated by documentation and release notes, not API enforcement.

### Alternatives Considered

| Approach | Rejected because |
|----------|-----------------|
| Keep CRD default + Go constant | Duplication, can drift |
| Mutating webhook | Operational complexity; DG spec comment explicitly avoids this |
| Make MaglevConfig required via CEL | Breaks minimal-config convenience |
| Non-pointer struct with +kubebuilder:default={} | Breaks has() CEL for future types; always serialized even when not applicable |
| Make MaxEndpoints a pointer | Unnecessary if the field is required within the block |

PR #164, opened 2026-06-09
Remove CRD default for maxEndpoints, require explicit value (#118)

zolug:
#118

MaxEndpoints must now be explicitly set when the maglev block is specified. The API server rejects `maglev: {}` without it. When the entire maglev block is omitted, the controller applies the built-in default of 102 (DefaultMaglevMaxEndpoints).

Changes:
- Remove +kubebuilder:default=32 from MaxEndpoints field
- Change DefaultMaglevMaxEndpoints from 32 to 102
- Remove duplicate defaultMaxEndpoints const from LB controller,
  use API-level constant consistently
- Remove misleading resizePolicy from GatewayConfiguration sample
- Regenerate CRD manifests
- Update documentation


bszirtes (approved):
Looks good to me.

PR #175, opened 2026-06-16
Pod cache label filtering (controller-manager)

### LINKED ISSUE #93: Add optional Pod label based filtering to reconcilers watching application Pods
Currently reconcilers receive events for every Pod in the namespace, then filter in the mapper function or in the reconcile function. The concern is that in a busy namespace with many unrelated Pods, this generates unnecessary reconcile evaluations.

**Gain depends on where in the pipeline the filtering happens:**

1. Cache-level filtering (label selector on informer) — biggest gain. Events for non-matching Pods never enter the cache, never trigger any processing. Might not usable here due to the "_OR problem_": 
Controller-runtime's cache.Options.ByObject applies a label selector at the informer level per GVK — meaning it filters ALL watches for that type across the entire manager. You can't have different label selectors for the same type in different controllers.

In our case, the manager has multiple Pod watches with different intent:
- **DG controller**: watches application Pods (would filter on e.g., meridio-2.nordix.org/managed=true)
- **ENC controller**: watches application Pods (same filter) AND watches SLLB Pods (filtered on gateway.networking.k8s.io/gateway-name)

If we set cache.ByObject[&corev1.Pod{}] with a label selector for the application Pod label, the cache would exclude SLLB Pods entirely — breaking the ENC controller's SLLB Pod watch. 
We'd need an OR condition: "has label meridio-2.nordix.org/managed OR has label gateway.networking.k8s.io/gateway-name". But cache.ByObject only accepts a single labels.Selector, which is AND-based (Kubernetes label selectors can express key exists with multiple requirements, but they're ANDed together — there's no OR operator in the label selector grammar).

*Upgrade concern with cache-level filtering:*

If the cache filter label were applied to LB Pods via the Deployment template, enabling it on upgrade creates a visibility gap:

  - Controller-manager restarts with cache filter enabled
  - Existing LB Pods don't have the label yet → invisible to the cache → controller-manager can't see them at all (not just events — List/Get from cache also returns nothing)
  - Controller-manager updates the Deployment template to include the label → triggers rolling update
  - During rollout: old Pods (no label) are invisible, new Pods (with label) are visible — a split-brain window where the controller-manager partially manages the LB fleet

Unlike predicate filtering where the cache holds everything and only event processing is skipped, cache-level filtering removes objects entirely from the controller-manager's view. 
There is no self-healing path — the controller-manager cannot reconcile what it cannot see.

2. **Predicate filtering** (`WithEventFilter` / `builder.WithPredicates`) — moderate gain. The event reaches the controller's event handler but is dropped before the mapper function runs. Avoids the mapper's API calls (listing DGs, matching selectors). This is a cheap label check vs. potentially multiple List calls in the mapper.

Predicates are per-watch, not per-type. So different predicates can be attached to different Watches calls.
The DG/Gateway/L34Route mappers are on their own Watches calls with no Pod label predicate. Those mappers receive DG/Gateway/Route events (not Pod events), then internally list Pods via the API client (which reads from the unfiltered cache). So they still find all Pods regardless of the label.

3. **Mapper filtering** (current approach) — the mapper runs for every Pod event, does List calls and selector matching, then returns empty if no match. This is the most expensive path per irrelevant event.

4. **In-reconcile filtering** — worst case. The work queue is populated, a reconcile is dequeued, and only then is the Pod found irrelevant.

**Conclusion:**

So predicate vs. mapper: the gain is real. For each irrelevant Pod event, you skip the mapper's List calls (DGs, selectors). In a namespace with 500 unrelated Pods and frequent Pod churn, that's a lot of avoided API calls. The predicate is just a label lookup on the object — O(1) vs. the mapper's O(n) list + match.

The gain is proportional to the ratio of unrelated Pods to managed Pods. If most Pods in the namespace are Meridio-managed, the benefit is minimal. If Meridio manages 10 Pods in a namespace with 500, it's significant.

*Operational considerations:*

- When enabling predicate the label filter on an existing deployment, application Pods without the label will stop receiving Pod-event-driven reconciliations. However, stale ENCs will self-heal:
DG/Gateway/L34Route mappers still enumerate all matching Pods from the unfiltered cache, so the next CR change triggers cleanup of ENCs for unlabeled Pods.
- The filter label configuration and its value should be documented as an operational tuning parameter, not a requirement. When unset, behavior is unchanged (all Pods processed).

zolug:
Fixes #93

## Background

Cache-level filtering was chosen over predicate-based filtering because:
- Both approaches require the same operational prerequisite (pre-labeling application Pods before enabling)
- Cache filtering provides strictly superior runtime performance: events for non-matching Pods never enter the informer, never allocate memory, never trigger any processing
- Predicate filtering still receives, deserializes, and evaluates every Pod event before dropping it
- The OR problem (DG and ENC controllers watching different Pod subsets) is resolved by using a single common label for both application and LB Pods
- When applied consistently, both approaches produce identical reconciliation and data-plane outcomes — cache filtering simply does so with less overhead

See #93 for the full analysis including per-controller impact, upgrade concerns, and runtime label-removal consequences.

## Changes

Pod cache label filtering (controller-manager):
- Add --pod-cache-label flag (MERIDIO_POD_CACHE_LABEL env) to scope the informer cache to Pods with a specified key=value label
- When set, only labeled Pods are visible to all controllers (DG, ENC, Gateway)
- Gateway controller stamps the label on LB Deployment pod templates automatically
- Feature is opt-in: empty value = disabled (all Pods cached, default behavior unchanged)
- Flag format validated in PreRunE (rejects malformed input)
- The LB and router controllers (running inside LB Pods) are not affected by the pod cache label — they do not watch Pod objects. The cache filtering only applies to the controller-manager, which hosts the DG and ENC controllers.

ENC controller improvements (discovered during testing):
- Fix non-deterministic ordering of GatewayConnections, NetworkDomains, and nextHops (caused update storms from map iteration order flapping)
- Handle 409 Conflict errors with immediate requeue instead of exponential backoff (faster convergence)
- Exclude LB Pods from .For(Pod) watch via predicate (avoids unnecessary reconciliation)
- Add FNV-32 hash at V(1) log level for compact change detection; full details at V(2)
- Refactor sort comparators to use cmp.Compare (also applied to bird and LB controller)

Example targets:
- Fix slow container termination: replace tail -f /dev/null with signal-aware trap pattern

E2E test suite (pod-cache-label):
- Verifies Gateway Accepted/Programmed status
- LB Pods inherit the cache label
- ENC created for labeled app Pod with non-empty next-hops
- ENC not created for unlabeled app Pod
- VIP ICMP reachability
- TCP traffic reaches only the labeled target

Documentation:
- docs/operations/pod-cache-filtering.md — configuration, requirements, upgrade procedure, operational constraints

## Refactoring

- `reconcileDeploymentSpec` converted to a method on GatewayReconciler (eliminates growing parameter list)
- `setPodCacheLabel` helper for label application with nil-safe map handling

### DIFF
--- a/cmd/controller-manager/cmd/run.go
+++ b/cmd/controller-manager/cmd/run.go
@@ -21,10 +21,12 @@ import (
 	"crypto/tls"
 	goflag "flag"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -87,6 +89,13 @@ func NewRunCmd() *cobra.Command {
 
 			if cfg.CertWaitTimeout > time.Minute {
 				return fmt.Errorf("--cert-wait-timeout cannot exceed 1 minute (got %s)", cfg.CertWaitTimeout)
+			}
+
+			if cfg.PodCacheLabel != "" {
+				k, v, ok := strings.Cut(cfg.PodCacheLabel, "=")
+				if !ok || k == "" || v == "" || strings.Contains(v, "=") {
+					return fmt.Errorf("--pod-cache-label must be in key=value format, got %q", cfg.PodCacheLabel)
+				}
 			}
 
 			return nil
@@ -183,6 +192,18 @@ func runManager(cfg *config.ManagerConfig) error {
 		}
 	}
 
+	// Parse pod cache label once (already validated in PreRunE)
+	var podCacheLabelKey, podCacheLabelValue string
+	if cfg.PodCacheLabel != "" {
+		podCacheLabelKey, podCacheLabelValue, _ = strings.Cut(cfg.PodCacheLabel, "=")
+		if cacheOptions.ByObject == nil {
+			cacheOptions.ByObject = map[client.Object]cache.ByObject{}
+		}
+		cacheOptions.ByObject[&corev1.Pod{}] = cache.ByObject{
+			Label: labels.SelectorFromSet(labels.Set{podCacheLabelKey: podCacheLabelValue}),
+		}
+	}
+
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                        scheme,
 		Cache:                         cacheOptions,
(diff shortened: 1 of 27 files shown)


ljkiraly (approved):
Looks good.

PR #184, opened 2026-06-25
E2E: add readiness gates test suite

### LINKED ISSUE #174: E2E;Readiness Gates; Verify end-to-end the three-step readiness gate feature
Follow-up issue of #123 relates to Readiness Gates issue: #90

Objective
Verify end-to-end that the three-step readiness gate feature works:

Gateway controller declares gates on LB Pods
Router sets gate conditions based on BGP state
ENC controller filters next-hops by gate conditions
Use or  reuse: dual-stack configuration (2 Gateways, 2 replicas each, BGP hold time 24s)

CI Integration
The tests use Serial + Ordered (same as resiliency tests) and live in e2e_suite_readiness_gates_test.go. They run after traffic tests.

Run with: --focus='Readiness Gates'

Test Cases
- Gates declared in spec and the status become True.
- ENC next-hops include all healthy LB IPs
- IPv4 BGP drop excludes Pod IPv4 RG status became False
- Link Flap don't change the RG status
- IPv6 BGP drop excludes Pod IPv6 RG status became False
- Link Flap don't change the RG status 
- Recover IPv4 and RG back to True Pods included as ENC next-hops
- Recover IPv6 and RG back to True Pods included as ENC next-hops

ljkiraly:
Verifies end-to-end the three-step readiness gate feature:
- Gateway controller declares per-IP-family gates on LB Pods
- Router sets gate conditions based on BGP state with damping
- ENC controller filters next-hops by gate conditions

Test cases cover per-family BGP drop/recovery, link flap damping, and ENC next-hop inclusion/exclusion. Uses dual-stack suite infrastructure (gw-ds, 2 replicas, holdTime 24s, BFD).

gh-174

### DIFF
--- a/test/e2e/Makefile
+++ b/test/e2e/Makefile
@@ -343,6 +343,23 @@ undeploy-dual-stack-simple: ## Undeploy dual-stack topology.
 	$(KUBECTL) delete validatingwebhookconfiguration meridio-2-validating-webhook-configuration-dual-stack --ignore-not-found=true; \
 	$(KUBECTL) delete namespace $$NS --ignore-not-found=true
 
+################################################################################
+##@ Suite: readiness-gates
+################################################################################
+
+.PHONY: readiness-gates
+readiness-gates: deploy-readiness-gates test-readiness-gates ## Deploy and test readiness-gates suite.
+
+.PHONY: deploy-readiness-gates
+deploy-readiness-gates: deploy-dual-stack-simple ## Deploy readiness-gates topology (reuses dual-stack-simple).
+
+.PHONY: test-readiness-gates
+test-readiness-gates: ## Run e2e readiness gates tests.
+	cd $(PROJECT_ROOT)/test/e2e && go run github.com/onsi/ginkgo/v2/ginkgo -v --tags=e2e --focus="Readiness Gates" $(GINKGO_FLAGS)
+
+.PHONY: undeploy-readiness-gates
+undeploy-readiness-gates: undeploy-dual-stack-simple ## Undeploy readiness-gates topology (same as dual-stack-simple).
+
 ###############################################################################
 ##@ All ipv4 suites
 ################################################################################
--- a/test/e2e/e2e_suite_readiness_gates_test.go
+++ b/test/e2e/e2e_suite_readiness_gates_test.go
@@ -28,6 +28,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	e2eutils "github.com/nordix/meridio-2/test/e2e/utils"
 	"github.com/nordix/meridio-2/test/utils"
 )
 
@@ -41,6 +42,8 @@ const (
 	rgProtocolV4  = "NBR-gw-ds-router-v4"
 	rgProtocolV6  = "NBR-gw-ds-router-v6"
 	rgTargetLabel = "app=target-ds"
+	rgVIPv4       = "10.0.0.1"
+	rgVIPv6       = "fd00:cafe:1::1"
 )
 
 var _ = Describe("Readiness Gates", Label("dual-stack"), Serial, Ordered, func() {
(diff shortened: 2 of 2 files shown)


ljkiraly:
[e2e](https://github.com/Nordix/Meridio-2/actions/workflows/e2e.yaml)
Job: https://github.com/Nordix/Meridio-2/actions/runs/28156463363

bszirtes (commented):
Great test, the ordered sequence of disable/verify/re-enable covers the readiness gate lifecycle well. A few suggestions:

#### Error handling in helpers

`rgGetGateCondition` and `rgGetTargetPod` discard the error from `utils.Run`. If kubectl fails transiently, the `Eventually`/`Expect` assertions will report confusing messages like `Expected "" to equal "False"` instead of
the actual kubectl error. Consider propagating or logging the error.

#### Cleanup safety net

Since the suite is `Ordered` and each test depends on state from the previous one, a mid-sequence failure (timeout, panic) leaves BGP protocols disabled with no recovery path. An `AfterAll` that unconditionally re-
enables both protocols would make this safe for repeated runs without redeployment:

```go
AfterAll(func() {
    rgBirdctl(lbPods[0], "enable", rgProtocolV4)
    rgBirdctl(lbPods[0], "enable", rgProtocolV6)
})
```

#### Coverage gaps worth considering

- Verify the Pod's overall `Ready` condition becomes `False` when both gates are `False` (and `True` when they recover). This validates the actual mechanism readiness gates are designed for.
- Assert that `lbPods[1]` gates remain True and its IPs stay in the ENC throughout the disruption — confirms the blast radius is limited to the disrupted pod.

#### Minor style

`rgGetSinglePodIP` uses `json.Unmarshal` directly while the rest of the file uses `utils.ParseJSON`. Small inconsistency.

ljkiraly:
Thanks for the suggestions. The corrections done by separate commits. 

The exception is the Pod's overall Ready condition check. In current feature (External Connectivity Exposure with Readiness Gates PR #90) readiness gates are used for connectivity-based readiness signaling. The Pod Ready status comes as a side effect and it is a kubernetes feature which is probably tested.

bszirtes (approved):
Looks good to me! Two optional enhancements to consider:

**Issue 1: ENC next-hops read without `Eventually` (line 117)**

```go
targetPod := rgGetTargetPod()
hops := rgGetENCNextHops(targetPod)
Expect(hops).To(ContainElements(ipv4IPs))
Expect(hops).To(ContainElements(ipv6IPs))
```

The ENC may not have been reconciled yet when this test runs immediately after `BeforeAll`. Every other ENC read in the suite uses `Eventually` — this one should too:

```go
targetPod := rgGetTargetPod()
Eventually(func() []string {
   return rgGetENCNextHops(targetPod)
}).WithTimeout(30  time.Second).WithPolling(2  time.Second).
   Should(ContainElements(ipv4IPs))
Eventually(func() []string {
   return rgGetENCNextHops(targetPod)
}).WithTimeout(30  time.Second).WithPolling(2  time.Second).
   Should(ContainElements(ipv6IPs))
```

**Issue 2: Error handling in helpers produces confusing failures**

`rgGetTargetPod` (line 368) and `rgGetGateCondition` (line 313) return `fmt.Sprintf("ERROR: %v", err)` as a string on kubectl failure. This gets passed to subsequent kubectl commands as a pod name or used in string comparisons, producing confusing error messages like `pod "ERROR: exit status 1" not found` instead of surfacing the original kubectl failure.

Consider using `ExpectWithOffset` to fail immediately at the call site (like `rgBirdctl` already does):

```go
func rgGetTargetPod() string {
   cmd := exec.Command("kubectl", "get", "pods", "-n", rgNamespace,
       "-l", rgTargetLabel, "-o", "jsonpath={.items[0].metadata.name}")
   out, err := utils.Run(cmd)
   ExpectWithOffset(1, err).NotTo(HaveOccurred(), "failed to get target pod: %s", out)
   return strings.TrimSpace(out)
}
```

bszirtes:
Looks good to me.

ljkiraly:
The mentioned issues are fixed in commit fa4aa63
For rgGetGateCondition, keep the error string pattern — it's used inside Eventually and the error message will show up in the assertion failure (e.g., Expected "ERROR: exit status 1" to equal "True") which is actually informative enough.
Verified by job: https://github.com/Nordix/Meridio-2/actions/runs/28586317556

Issue #103, opened 2026-04-14
Static+BFD: CRD and bird config generation


ljkiraly:
Update CRD with Static fields.
Create bird configuration from custom resource.

zolug:
A candidate GatewayRouter CRD could look like this:

```go
// GatewayRouterSpec defines the desired state of GatewayRouter
//
// +kubebuilder:validation:XValidation:rule="self.protocol == 'BGP' ? has(self.bgp) && !has(self.static) : has(self.static) && !has(self.bgp)",message="If protocol is BGP, bgp must be set and static must not be set. If protocol is Static, static must be set and bgp must not be set."
type GatewayRouterSpec struct {
	// gatewayRef references the Gateway this router peers with.
	GatewayRef gatewayapiv1.ParentReference `json:"gatewayRef"`

	// interface is the name of the network interface used to reach the external gateway.
	Interface string `json:"interface"`

	// +kubebuilder:validation:XValidation:rule=isIP(self),message=Must be an ip address
	// address is the IP address of the external Gateway Router.
	Address string `json:"address"`

	// protocol selects the routing protocol for this peering.
	// +kubebuilder:validation:Enum=BGP;Static
	Protocol RoutingProtocol `json:"protocol"`

	// bgp defines BGP session parameters. Required when protocol is BGP.
	// +optional
	BGP *BgpSpec `json:"bgp,omitempty"`

	// static defines static routing with BFD supervision. Required when protocol is Static.
	// +optional
	Static *StaticSpec `json:"static,omitempty"`
}

// +enum
type RoutingProtocol string

const (
	RoutingProtocolBGP    RoutingProtocol = "BGP"
	RoutingProtocolStatic RoutingProtocol = "Static"
)
```

- Discriminator field protocol with enum BGP/Static. Immutable after creation would be ideal.
- Mutual exclusivity via CEL at the spec level: one of bgp/static must be set, matching the protocol value.
- BGP changed from required struct to optional pointer (*BgpSpec). This is a "breaking change" from the current CRD where bgp is a required non-pointer field. Necessary to support the "absent when not selected" pattern. (No backward compatibility requirements yet towards baseline.)
- BfdSpec reused across both modes.

elajkat:
some additional details, which can be used for the actual design, or help in the final clarification also:

- As we anyway have Address field for the GatewayRouter we can use the same, so a minimal StaticSpec can be:
```
type StaticSpec struct {
    // BFD monitoring of the static next-hop.
[...]

Issue #104, opened 2026-04-14
Static+BFD: Monitoring and external connectivity exposure


ljkiraly:
Monitor BFD status and expose it.
depends on External Connectivity Exposure via Readiness Gates #90

PR #188, opened 2026-07-01
Add crd, and bird config generation for static routing

arp-est:
https://github.com/Nordix/Meridio-2/issues/103


arp-est:
There was one thing that bugged, me,
It's when you set the bfd for the static, and for that you need to put it to the global bfd section, and its defined for an interface.
So if multiple gatewayrouter cr-s are static and share the same interface you can't determine which ones paramters should win. According to ai in meridio-1 the static route that was first specified wins, so I went with that.

I should probably make a test for it.

EDIT:
The gateway routers are sorted alphabetically and the first one in this order will have its parameters set if they compete for the same interface. (Also added a test)

elajkat:
I cant add comment for files not touched by this PR (I hate github....):
in bird/monitor.go:
// IsEstablished checks if the protocol has an established BGP session
// For BGP protocols, both State must be "up" AND Info must contain "Established"
func (p ProtocolStatus) IsEstablished() bool {
	return p.State.IsUp() && strings.Contains(p.Info, "Established")
}
for static there is no established in the protocol show output:
$ kubectl exec -n e2e-static-bfd sllbr-gw-s1-7dff4499c9-rvh78 -c router -- birdc -s /var/run/bird/bird.ctl show protocols 2>&1
BIRD 3.2.2 ready.
Name       Proto      Table      State  Since         Info
NBR-gw-s1-router-v4 Static     master4    up     12:09:35.224  

so Proto is Static, State is up and Info is empty (at least in my env)

So I suppose something like this can work with a switch:
switch p.Proto {
   case "Static": 
      return p.State.IsUp()                                                                                                                                                            
   default:                                                                                                                                                                           
      return p.State.IsUp() && strings.Contains(p.Info, "Established")
}

zolug:
> I cant add comment for files not touched by this PR (I hate github....): in bird/monitor.go: // IsEstablished checks if the protocol has an established BGP session // For BGP protocols, both State must be "up" AND Info must contain "Established" func (p ProtocolStatus) IsEstablished() bool { return p.State.IsUp() && strings.Contains(p.Info, "Established") } for static there is no established in the protocol show output: $ kubectl exec -n e2e-static-bfd sllbr-gw-s1-7dff4499c9-rvh78 -c router -- birdc -s /var/run/bird/bird.ctl show protocols 2>&1 BIRD 3.2.2 ready. Name Proto Table State Since Info NBR-gw-s1-router-v4 Static master4 up 12:09:35.224
> 
> so Proto is Static, State is up and Info is empty (at least in my env)
> 
> So I suppose something like this can work with a switch: switch p.Proto { case "Static": return p.State.IsUp() default: return p.State.IsUp() && strings.Contains(p.Info, "Established") }

Integrating with monitoring shall come in a separate issue: https://github.com/Nordix/Meridio-2/issues/104

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -33,10 +35,20 @@ type GatewayRouterSpec struct {
...
 
+	// protocol selects the routing protocol for this peering.
+	// +kubebuilder:validation:Enum=BGP;Static
+	Protocol RoutingProtocol `json:"protocol"`
+
 	// Parameters to set up the BGP session to specified Address.
 	// If the Protocol is bgp, the minimal parameters to be defined in bgp properties
 	// are RemoteASN and LocalASN
+	// +optional
 	BGP BgpSpec `json:"bgp"`
```
zolug: please check this comment about BgpSpec: https://github.com/Nordix/Meridio-2/issues/103#issuecomment-4378640555

internal/bird/bird.go
```diff
@@ -177,15 +177,37 @@ func (b *Bird) generateConfig(vips []string, routers []*meridio2v1alpha1.Gateway
...
 
 	ifset := make(map[string]bool)
+	bfdParams := make(map[string]*meridio2v1alpha1.BfdSpec)
 	for _, r := range routers {
-		rd, err := toRouterData(r)
-		if err != nil {
-			return "", err
-		}
-		data.Routers = append(data.Routers, rd)
 		ifset[r.Spec.Interface] = true
```
zolug: Shouldn't this ifset operation come after the switch block? (It used to be preceded by routing data transformation and error handling)

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -22,6 +22,8 @@ import (
 )
 
 // GatewayRouterSpec defines the desired state of GatewayRouter
+// +kubebuilder:validation:XValidation:rule="self.protocol != 'Static' || has(self.static)",message="static is required when protocol is Static"
+// +kubebuilder:validation:XValidation:rule="self.protocol != 'BGP' || (has(self.bgp) && self.bgp.remoteASN > 0 && self.bgp.localASN > 0)",message="bgp.remoteASN and bgp.localASN are required when protocol is BGP"
 type GatewayRouterSpec struct {
```
zolug, line 28: missing mutual exclusivity validation (refer to https://github.com/Nordix/Meridio-2/issues/103#issuecomment-4378640555)

Can be expressed in CEL once BGP becomes a pointer.

internal/bird/bird.go
```diff
@@ -177,15 +177,37 @@ func (b *Bird) generateConfig(vips []string, routers []*meridio2v1alpha1.Gateway
...
+
+		switch r.Spec.Protocol {
+		case meridio2v1alpha1.RoutingProtocolStatic:
+			data.StaticRouters = append(data.StaticRouters, toStaticRouterData(r))
+			if _, exists := bfdParams[r.Spec.Interface]; !exists && isStaticBFDOn(r) {
```
zolug, line 186: Documentation should describe this first one alphabetically wins concept. Also, similarly to Meridio v1 documentation should mandate users to define the same BFD params for static routers on the same interface.

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -77,6 +89,20 @@ type BgpSpec struct {
...
+	RoutingProtocolStatic RoutingProtocol = "Static"
+)
+
+type StaticSpec struct {
+	// BFD monitoring of the static next-hop.
+	// +optional
+	BFD *BfdSpec `json:"bfd,omitempty"`
+}
+
 type BfdSpec struct {
```
zolug, line 113: Meridio v1 used a webhook that seemingly required MinTx, minRx and multiplier values to be set explicitly in case of switch=true. I think we should mimic the same v1 behaviour (i.e. to make them mandatory when enabled) instead of deferring defaulting to BIRD. Any thoughts?

So we could have sg like this via CEL on BfdSpec:
`// +kubebuilder:validation:XValidation:rule="!has(self.switch) || self.switch == false || (self.minTx != '' && self.minRx != '' && has(self.multiplier))",message="minTx, minRx and multiplier are required when switch is true"`

arp-est: how about I just make the fields required, if BFD is not nil?, and the switch, can just mean to add it to bird config, or not.
so something like:

```go
type BfdSpec struct {
	// BFD monitoring.
	// Valid values are:
	// - false: no BFD monitoring;
	// - true: turns on the BFD monitoring.
	// +required
	Switch *bool `json:"switch,omitempty"`

	// Min-tx timer of bfd session. Please refere to BFD material to understand what this implies.
	// The value must be a valid duration format. For example, 300ms, 90s, 1m, 1h.
	// The duration will be rounded by millisecond.
	// +required
	MinTx string `json:"minTx,omitempty"`

	// Min-rx timer of bfd session. Please refere to BFD material to understand what this implies.
	// The value must be a valid duration format. For example, 300ms, 90s, 1m, 1h.
	// The duration will be rounded by millisecond.
	// +required
	MinRx string `json:"minRx,omitempty"`

	// Multiplier of bfd session.
	// When this number of bfd packets failed to receive, bfd session will go down.
	// +required
	Multiplier *uint16 `json:"multiplier,omitempty"`
}

```

zolug: > how about I just make the fields required, if BFD is not nil?, and the switch, can just mean to add it to bird config, or not. so something like:
> 
> ```go
> type BfdSpec struct {
> 	// BFD monitoring.
> 	// Valid values are:
> 	// - false: no BFD monitoring;
> 	// - true: turns on the BFD monitoring.
> 	// +required
> 	Switch *bool `json:"switch,omitempty"`
> 
> 	// Min-tx timer of bfd session. Please refere to BFD material to understand what this implies.
> 	// The value must be a valid duration format. For example, 300ms, 90s, 1m, 1h.
> 	// The duration will be rounded by millisecond.
> 	// +required
> 	MinTx string `json:"minTx,omitempty"`
> 
> 	// Min-rx timer of bfd session. Please refere to BFD material to understand what this implies.
> 	// The value must be a valid duration format. For example, 300ms, 90s, 1m, 1h.
> 	// The duration will be rounded by millisecond.
> 	// +required
> 	MinRx string `json:"minRx,omitempty"`
> 
> 	// Multiplier of bfd session.
> 	// When this number of bfd packets failed to receive, bfd session will go down.
> 	// +required
> 	Multiplier *uint16 `json:"multiplier,omitempty"`
> }
> ```

Due to the Switch field that would require a user to set minTx and the other values even if Switch=false, which would be odd. Also, I don't recall why in v1 we had this knob based approach (i.e. dedicated switch).

arp-est: For the record, removed the switch field, now "bfd on" is dependent if the struct is filled and not nil.
Also changed multiplier to be a regular uint16 instead of pointer, as with all fields required there's no need for it to be a pointer.

internal/bird/config.go
ljkiraly, line 91: The `gateway_routes` filter (used for kernel export) currently accepts only `RTS_BGP`
Static routes get imported via `default_rt` filter into the main routing table, then need to be exported to the kernel table. The default route (`0.0.0.0/0` / `0::/0`) IS explicitly accepted, so static default routes DO pass through. However, this means only default routes work for static. If in the future static routes beyond default are added, they'd be silently dropped.
Looking at the Meridio v1 reference config in the issue, v1 has a separate `cluster_breakout` filter that accepts both `RTS_BGP` and (implicitly via the default route match) static routes. The current approach works for the default-route-only use case, but adding `if source = RTS_STATIC then accept;` to `gateway_routes` would be more robust and future-proof.

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -33,17 +37,31 @@ type GatewayRouterSpec struct {
 	// Address of the Gateway Router
 	Address string `json:"address"`
 
+	// protocol selects the routing protocol for this peering.
+	// +kubebuilder:validation:Enum=BGP;Static
+	Protocol RoutingProtocol `json:"protocol"`
```
ljkiraly, line 43: `protocol` field itself is not immutable. 

The `bgp` and `static` fields are individually marked immutable once set, but the `protocol` field has no immutability constraint. A user could change `protocol: BGP` to `protocol: Static`, so the BGP field remains (it's immutable, can't be removed), creating an inconsistent state that the mutual exclusivity CEL would reject. This may be fine (CEL catches it), but the error message would be confusing. Consider adding:
```
// +kubebuilder:validation:XValidation:rule="self.protocol == oldSelf.protocol",message="protocol is immutable"
```
on the spec, or document that protocol changes are blocked by the mutual exclusivity rules.

test/e2e/suites/dual-stack/routing.yaml
ljkiraly, line 13: *minor* Consider adding `protocol: BGP` for consistencty with other e2e tests.

test/e2e/suites/dual-stack/routing.yaml
```diff
@@ -16,7 +16,6 @@ spec:
     remotePort: 10179
     holdTime: 24s
     bfd:
```
ljkiraly, line 19: *minor* Consider adding `protocol: BGP` for consistencty with other e2e tests.

test/e2e/suites/dual-stack/routing.yaml
ljkiraly, line 35: *minor* Consider adding `protocol: BGP` for consistencty with other e2e tests.

arp-est: I thought I've added them I guess I missed this file, thanks!

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -77,31 +95,37 @@ type BgpSpec struct {
...
+	// Min-tx timer of bfd session. Please refer to bird BFD documentation to understand what this implies.
 	// The value must be a valid duration format. For example, 300ms, 90s, 1m, 1h.
 	// The duration will be rounded by millisecond.
-	// +optional
-	MinTx string `json:"minTx,omitempty"`
+	// +required
+	MinTx string `json:"minTx"`
 
-	// Min-rx timer of bfd session. Please refere to BFD material to understand what this implies.
+	// Min-rx timer of bfd session. Please refere to bird BFD documentation to understand what this implies.
```
ljkiraly: *minor*: typo in comment: "Please refere to bird BFD documentation" → "Please refer to bird BFD documentation"

ljkiraly (commented):
Verdict: Can be approved with suggestions.

### Architecture & Patterns
- CRD design follows the recommended pattern
- Must-fix: The gateway_routes filter doesn't include RTS_STATIC acceptance. While default routes pass through via the 0.0.0.0/0 match, this is fragile if non-default static routes are ever needed. Adding `if source = RTS_STATIC && dest != RTD_BLACKHOLE then accept;` (as Meridio v1 did) future-proofs it.

### Technical Correctness
1. Code correctness
-  Design gap: The protocol field itself has no immutability constraint (only bgp/static do). Changing protocol is caught by mutual exclusivity CEL but produces confusing error messages.
2. Security - No security concerns.
3. Error handling - Good.
4. Integration safety 
- The `filter default_rt` addition is unconditional (always emitted even without static routers) - harmless
### Test Coverage & Maintainability
- Good coverage
- No test for the `""` (empty) protocol backward compatibility path. Would be good to have one test with a router without protocol field to ensure it still generates BGP config
- **Test refactoring:** The move from nested closures to named functions (`testStaticRouterWithBFD`, etc.) improves readability. Good pattern.
- **Documentation:** Well-updated in both `router.md` and `constraints-and-limitations.md`.

### Recommendations

#### Must-fix before merge

1. **Add `RTS_STATIC` to `gateway_routes` filter** (or document why it's not needed). Even though default routes are already matched by the `0.0.0.0/0` rule, adding `if source = RTS_STATIC && dest != RTD_BLACKHOLE then accept;` (matching v1's `cluster_access` pattern) future-proofs the filter and makes intent explicit.

#### Nice-to-have / follow-up

2. **Add `protocol` field immutability CEL** — prevents confusing error messages when a user tries to switch protocols.

3. **Add a backward-compatibility test** for empty `protocol` field → treated as BGP.

4. **Consider adding `protocol: BGP`** to `dual-stack/routing.yaml` for consistency with other e2e tests (currently relies on the empty-string fallback).

5. **BFD template formatting** — the inline format `{min rx interval 300ms; min tx interval 300ms; multiplier 3;}` is valid BIRD syntax but less readable than multi-line. Minor style point.

6. **Typo in BfdSpec comment**: "Please refere to bird BFD documentation" → "Please refer to bird BFD documentation" (line 109 in types.go).

internal/bird/config.go
ljkiraly: The `gateway_routes` filter is for kernel export (what routes to install into the kernel routing table), and default routes (0.0.0.0/0, 0::/0)  are already explicitly accepted there. Static default routes pass through fine. The `RTS_STATIC` check in v1 is on the export-to-BGP-peers path, which v2 already handles via `announced_routes`. 

No change needed to gateway_routes.

ljkiraly (approved):
Looks good.

PR #190, opened 2026-07-03
Extract controller manager into pkg/app for downstream extensibility

### LINKED ISSUE #192: Implement controller manager extension mechanism
## Summary

Extract the controller manager into an exported `pkg/app` package enabling downstream consumers to register additional controllers in the same manager instance (shared informer cache, shared leader election, single binary).

## Design Decisions

- Package location: `pkg/app`
- Entry point: `NewCommand(additional ...ControllerSetup)` returns a Cobra command with all flags/env vars
- `ControllerSetup func(mgr ctrl.Manager, cfg Config) error` — receives resolved config view
- `Config` struct exposes `Namespace` and `ControllerName` (resolved from flags > env > defaults)
- Config type stays in `internal/` — downstream uses `NewCommand()` and never imports it directly
- Scheme/webhook registration handled internally
- Leader election ID shared (documented constraint)

## Implementation

- `pkg/app/app.go` — `NewCommand()`, `ControllerSetup`, `Config`
- `cmd/controller-manager/cmd/run.go` — thin wrapper: `app.NewCommand()`
- `build/controller-manager/Dockerfile` — add `COPY pkg/ pkg/`
- `pkg/app/README.md` — exported API, downstream usage, constraints
- Unit tests for validation and command creation

## Related

- Study: #133
- PR: #190

ljkiraly:
Introduce pkg/app with NewCommand() for embedding Meridio-2 controllers in downstream binaries. NewCommand() provides same flags and env var bindings as the default binary — downstream consumers never import internal packages. Additional ControllerSetup functions share the informer cache and leader election.

cmd/controller-manager/cmd/run.go becomes a thin wrapper delegating to app.NewCommand(). No behavioral change for default deployments.

Fixes: gh-192


alitta-german:
Reviewed the implementation — this is exactly what we need for our downstream EgressAggregator controller. The API design is clean and the NewCommand() + Config{Namespace, ControllerName} pattern is actually better than what we originally proposed in our extension proposal (we had Run(cfg, additional...) which would have required downstream config construction).

  Our SetupWithManager adapts trivially to func(ctrl.Manager, app.Config) error. We'll use cfg.Namespace directly instead of self-parsing it.

  No blocking concerns from our side. Looking forward to a tagged release so we can import pkg/app as a Go module dependency.

PR #194, opened 2026-07-10
Replace EndpointSlice with custom LoadBalancerEndpointSlice CRD

zolug:
Implements the DG controller (producer) side of #178.

### Summary

Replaces Kubernetes core `EndpointSlice` with a purpose-built `LoadBalancerEndpointSlice` CRD for endpoint discovery between the DG controller and LB controller. This eliminates the abuse of core API fields (Zone for Maglev IDs, per-CIDR label encoding) and provides a clean, typed contract.

### Key design changes

- **Unified dual-stack model**: One endpoint entry per Pod carries all addresses (IPv4 + IPv6) — no cross-object correlation needed by consumers
- **Per-Gateway scoping**: Each slice has `spec.gatewayRef` identifying its Gateway (replaces implicit per-CIDR association)
- **Direct Maglev ID field**: `spec.endpoints[].identifier` replaces the `zone: "maglev:<id>"` encoding
- **Pod identity**: `spec.endpoints[].target` carries Pod name + UID for staleness detection
- **Deterministic naming**: Slice names use `<dg-name>-<FNV64a-hash>-<index>` (hash includes DG + Gateway identity, future-proofs for multi-Gateway)

### Changes

**CRD definition:**
- `api/v1alpha1/loadbalancerendpointslice_types.go` — full type with kubebuilder markers
- Generated CRD YAML + deepcopy
- Added to `config/crd/kustomization.yaml`
- Design rationale: `docs/design/meridio2-endpoint-producer-consumer-architecture.md`

**DG controller migration:**
- `slices.go` — rewritten for new model (scraping, slice building, compaction, structure preservation)
- `maglev.go` — reads `Identifier` field directly (no zone parsing)
- `helpers.go` — added `sliceBaseName` (FNV-64a hash), `truncateLabelValue`; removed CIDR encode/decode
- `const.go` — updated labels and constants
- `types.go` — `podWithAddresses` replaces `podWithNetworkIP`
- `status.go` — simplified capacity reporting (no longer per-network)
- `controller.go` — `Owns(&LoadBalancerEndpointSlice{})`, `MaxEndpointsPerSlice` field

**Configuration:**
- `--max-endpoints-per-slice` flag (default 200) + `MERIDIO_MAX_ENDPOINTS_PER_SLICE` env
- Passed from config to reconciler

**RBAC:**
- Replaced `discovery.k8s.io/endpointslices` with `meridio-2.nordix.org/loadbalancerendpointslices` in manager Role

**Tests:**
[...]

Issue #178, opened 2026-06-18
Replace EndpointSlice with custom LoadBalancerEndpointSlice CRD


zolug:
## Description

Replace the current Kubernetes EndpointSlice-based endpoint representation with a purpose-built `LoadBalancerEndpointSlice` custom resource. This eliminates schema limitations of the core EndpointSlice type and provides a clean, explicit contract between the DG controller (producer) and LB controller (consumer).

No backward compatibility with the current EndpointSlice approach is required.

## Motivation

The current EndpointSlice-based approach has fundamental limitations:
- `addressType` forces separate slices per IP family (IPv4/IPv6) — requires cross-slice correlation for dual-stack
- `Zone` field abused for Maglev IDs — no explicit field, string parsing required
- No Gateway association on slices — no way to scope endpoints per Gateway
- Label-based discovery subject to 63-character value limit
- No distribution type context — LB must infer from DG spec

## Proposed Schema

```yaml
apiVersion: meridio-2.nordix.org/v1alpha1
kind: LoadBalancerEndpointSlice
metadata:
  name: dg-1-0                    # <dg>-<index> (capacity splitting)
  namespace: default
  ownerReferences:
    - apiVersion: meridio-2.nordix.org/v1alpha1
      kind: DistributionGroup
      name: dg-1
      controller: true
spec:
  distributionGroup: dg-1
  gatewayRef:
    name: gw-a
    namespace: default
  distributionType: Maglev         # see open question below
  endpoints:
    - target:
        name: app-pod-1
        uid: abc-123
      addresses:
        - ip: "192.168.1.10"
          family: IPv4
        - ip: "2001:db8::10"
          family: IPv6
      identifier: 5
      ready: true
```

### Key design decisions

| Field | Rationale |
|-------|-----------|
| `spec.distributionGroup` | Plain string (DG name). Same-namespace relationship enforced by ownerReference — no namespace field needed. Supports native field selectors and `kubectl get --field-selector`. Follows core K8s EndpointSlice pattern (service name as label). |
| `spec.gatewayRef` | Exclusive per-Gateway; enables per-Gateway scoping. Separate slices per Gateway for future multi-GW |
| `spec.distributionType` | See open question below |
| `endpoints[].target` | Typed struct (name + uid) for debugging/correlation |
| `endpoints[].addresses` | Dual-stack in one object — no cross-slice correlation needed |
| `endpoints[].addresses[].family` | LB knows which routing table family to use |
[...]

PR #198, opened 2026-07-13
Migrate LB controller to LoadBalancerEndpointSlice

zolug:
Consumer-side migration for #178. Switches the stateless-load-balancer controller from Kubernetes core `EndpointSlice` to the custom `LoadBalancerEndpointSlice` CRD.

Depends on: #194 (DG controller / producer side)

### Summary

The LB controller now reads endpoint data directly from typed fields instead of parsing Zone strings and correlating across per-family slices.

### Key changes

**Production code:**
- `controller.go` — Watch `LoadBalancerEndpointSlice` instead of `discoveryv1.EndpointSlice`, register field indexer for `spec.distributionGroupName`
- `targets.go` — Read `Identifier` field directly (no Zone parsing), `Ready` as bool, dual-stack IPs from single endpoint, first-occurrence-wins for duplicate identifiers, ownerRef verification
- `flows.go` — `hasReadyEndpoints` rewritten with field index lookup + ownerRef check

**Design decisions:**
- **Field indexer + ownerRef verification**: Uses `client.MatchingFields` for efficient lookup, then verifies ownerReference as defense-in-depth against manually-created slices
- **First-occurrence-wins**: During transients, the same identifier may briefly appear in multiple slices; sorted-first slice takes priority (aligns with DG controller's compaction direction)
- **No IP accumulation across slices**: Unlike the old model (IPv4 + IPv6 in separate slices), dual-stack is now in a single endpoint entry — no cross-slice correlation needed

**RBAC:**
- LB Pod Role: replaced `discovery.k8s.io/endpointslices` with `meridio-2.nordix.org/loadbalancerendpointslices`

**Tests:**
- All `reconcileTargets` tests rewritten for new model
- `reconcileFlows` tests updated with field indexer on fake client
- `endpointSliceEnqueue` tests migrated to `LoadBalancerEndpointSlice`
- Removed obsolete Zone-parsing and cross-slice accumulation tests
- Added dual-stack single-endpoint test

### CRD cherry-pick

Includes the `LoadBalancerEndpointSlice` CRD definition (cherry-picked from #194) as a build dependency. The CRD commit is shared between both PRs.

### Testing

- `go build ./...` ✅
- `go test ./internal/controller/loadbalancer/` ✅ (30/30 pass)
- `make lint` ✅
- Manual deployment verified with DG controller producing slices


api/v1alpha1/loadbalancerendpointslice_types.go
```diff
@@ -0,0 +1,150 @@
...
+// LoadBalancerEndpointSliceSpec defines the desired state of LoadBalancerEndpointSlice.
+type LoadBalancerEndpointSliceSpec struct {
+
+	// distributionGroupName is the name of the DistributionGroup that owns this slice.
+	// Same-namespace relationship is enforced by ownerReference.
+	// +kubebuilder:validation:MinLength=1
+	DistributionGroupName string `json:"distributionGroupName"`
+
+	// gatewayRef identifies the Gateway this slice is scoped to.
+	// Enables per-Gateway endpoint discovery by the LB controller.
```
ljkiraly, line 32: Minor AI review suggestion:
### One Observation

The `gatewayRef` field is described as "Enables per-Gateway endpoint discovery by the LB controller" but the LB controller does **not** filter by it — it only filters by `distributionGroupName` (field index) and ownerReference. The LB relies on the higher-level `belongsToGateway()` check to ensure it only processes DGs for its Gateway, which transitively ensures it only sees slices relevant to its Gateway. This is correct and sufficient because:
1. Each DistributionGroup is scoped to exactly one Gateway (enforced by the DG controller's single-Gateway restriction)
2. The LB controller only reconciles DGs that pass `belongsToGateway()`
3. Slices owned by a DG are inherently scoped to that DG's Gateway

However, the field comment is slightly misleading — `gatewayRef` serves more as metadata/observability than as a filtering mechanism for the LB. This is fine for v1alpha1 but worth clarifying in comments.

ljkiraly (approved):
The code is correct and well-tested.

Consider a minor documentation update in this PR or the other one: the `docs/controllers/distributiongroup.md` context entry mentions "EndpointSlice" as the backend type in the architecture table. It should be updated to mention `LoadBalancerEndpointSlice`.

Some future task related to this change, which can be covered in follow-up issues:
1/ Missing test: duplicate identifier across slices: The "first-occurrence-wins" logic is documented but not tested. A test with two slices both containing `identifier=0` would validate the sorting + skip behavior.
2/ **Consider adding `gatewayRef` to the field indexer** — While not needed today (single DG → single Gateway), if future scenarios involve shared DGs across Gateways, the LB could use `client.MatchingFields{"spec.gatewayRef.name": c.GatewayName}` for more efficient filtering.

api/v1alpha1/loadbalancerendpointslice_types.go
```diff
@@ -0,0 +1,150 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// LoadBalancerEndpointSliceSpec defines the desired state of LoadBalancerEndpointSlice.
+type LoadBalancerEndpointSliceSpec struct {
+
+	// distributionGroupName is the name of the DistributionGroup that owns this slice.
+	// Same-namespace relationship is enforced by ownerReference.
+	// +kubebuilder:validation:MinLength=1
```
ljkiraly, line 28: Be careful when merging: the 'lb-slice' branch has the XValidation:rule="self == oldSelf" immutability validation, added in a later commit.

zolug:
> Consider a minor documentation update in this PR or the other one: the docs/controllers/distributiongroup.md context entry mentions "EndpointSlice" as the backend type in the architecture table. It should be updated to mention LoadBalancerEndpointSlice.

> Be careful when merging: the 'lb-slice' branch has the XValidation:rule="self == oldSelf" immutability validation, added in a later commit.

This PR only contains the LB part except for the CRD changes, and is solely meant for easing the review process.

Issue #196, opened 2026-07-13
Implement dynamic log level changes at runtime


bszirtes:
This issue is a follow-up task for #179, based on downstream consumer feedback.

## Summary

Implement runtime log level changes for all Meridio-2 components without requiring container/Pod restarts. This feature enables ad-hoc debugging and operational flexibility by exposing `zap.AtomicLevel` via a localhost-only HTTP endpoint.

## Background

Currently, log levels in Meridio-2 are set at startup via `--log-level` flag or `MERIDIO_LOG_LEVEL` environment variable and cannot be changed without restarting the container. This limitation makes debugging difficult in production environments where Pod restarts cause service disruption.

### Functional Requirements

1. **Runtime log level modification** - Change log level without Pod/container restart
2. **Per-Pod targeting** - Change level on specific Pods for ad-hoc debugging
3. **Multiple log levels** - Support arbitrary levels (debug, info, warn, error), not just toggle
4. **Query current level** - Ability to inspect current log level via API
5. **Fail-safe operation** - Feature failure must not prevent component startup
6. **Opt-in by default** - No behavioral change for existing deployments unless explicitly enabled

### Non-Functional Requirements

1. **Security** - Endpoint must bind to localhost only (no network exposure)
2. **Minimal code** - Leverage built-in zap.AtomicLevel HTTP handler (~10 lines per component)
3. **No external dependencies** - Use only stdlib and existing controller-runtime/zap
4. **Zero-config default** - Feature disabled when not configured (no ports opened)
5. **Ephemeral changes** - Level changes are intentionally not persisted (restart resets to default)

### Non-Goals

- Fleet-wide coordination (users can script per-Pod changes if needed)
- Persistence across restarts (use `--log-level` flag/env var for permanent changes)
- ConfigMap/CRD-based control (keep components Kubernetes-agnostic)
- Authentication/authorization (localhost-only binding provides security boundary)

### Option 3: HTTP Endpoint with zap.AtomicLevel

Expose `zap.AtomicLevel` via HTTP on a localhost-only address. This is the recommended approach based on:

- **Downstream consumer validation** (SENP2 team endorsement)
- **Ecosystem precedent** (cert-manager, Kubernetes controller-manager use this pattern)
- **Minimal implementation** (zap provides built-in HTTP handler)

#### 1. Core Helper Package

Create `internal/common/log/dynamic.go`:

```go
package log

[...]

PR #199, opened 2026-07-14
Add dynamic log level change API for runtime debugging

bszirtes:
## Summary

Implement HTTP endpoints on all Meridio-2 components to enable runtime log level changes without requiring pod restarts. This feature addresses the critical requirement for per-container log level control in sidecar deployments where application teams control their pods and Meridio-2 components are injected as sidecars.

## Issue

#196 

## Motivation

Downstream consumers require the ability to change log levels at runtime for debugging purposes without restarting pods. This is particularly critical for the Network Sidecar, which runs in application pods that Meridio-2 operators don't control. The existing approach of restarting pods disrupts application workloads and is not acceptable in production environments.

## Features

- **HTTP API using zap.AtomicLevel** - Standard zap library endpoint (GET/PUT `/log/level`)
- **Opt-in by default** - Disabled unless `MERIDIO_LOG_LEVEL_API` environment variable is set
- **Localhost-only binding** - Enforced `127.0.0.1` with security validation and fail-safe behavior
- **Non-blocking startup** - HTTP server runs in goroutine, errors are logged but non-fatal
- **Ephemeral changes** - Log levels reset on container restart (by design)
- **Multi-container safe** - Different ports prevent conflicts (9901/9902/9903)
- **Per-container granularity** - Each container has independent log level control

### New Package: `internal/common/log`

- `StartDynamicLevelServer()` - Starts HTTP server with security validation
- `ParseLevel()` - Validates log level strings with proper error handling
- Comprehensive unit tests with 96% coverage

### Configuration

Added `LogLevelAPI` field to all component configs:
- Flag: `--log-level-api`
- Environment variable: `MERIDIO_LOG_LEVEL_API`
- Default: empty (feature disabled)

### Cleanup: removed unused `--zap-*` flags

As part of wiring the atomic log level into each binary's `zap.Options`, the
kubebuilder-scaffolded `--zap-*` flags (bound via `zapOpts.BindFlags`) were
removed. These flags were never documented or referenced anywhere in the
project, so this is a cleanup rather than a breaking change.

### Security

- **Address validation** - Only loopback addresses (127.0.0.1, ::1) are accepted
- **Fail-safe behavior** - Server not started if validation fails
- **Security audit logging** - Violations are logged for audit trails
- **Attack prevention** - ReadHeaderTimeout set to prevent slowloris attacks
- **Network isolation** - Not reachable from outside the pod

### Get Current Log Level
```bash
curl http://127.0.0.1:9901/log/level
# Response: {"level":"info"}
```

### Change Log Level
```bash
curl -X PUT http://127.0.0.1:9901/log/level -d 'level=debug'
# Response: {"level":"debug"}
```

### Valid Log Levels
- `debug`
- `info`
- `warn`
- `error`

### Deployment Configuration

**Controller Manager Example** (via Kustomize):
```yaml
env:
- name: MERIDIO_LOG_LEVEL
 value: info
- name: MERIDIO_LOG_LEVEL_API
 value: 127.0.0.1:9901
```

**Network Sidecar Example** (in target pods):
```yaml
env:
- name: MERIDIO_LOG_LEVEL_API
 value: 127.0.0.1:9903
```

## Sidecar Use Case

The localhost HTTP endpoint enables per-container control in multi-container pods. A platform-specific glue layer running in the same pod can:

1. Read log level from a ConfigMap (platform-specific)
2. Invoke the HTTP endpoint: `curl -X PUT http://127.0.0.1:9903/log/level -d 'level=debug'`
3. Change only the Meridio-2 sidecar's log level without affecting the application container
4. No pod restart required

This satisfies the requirement for injected sidecars in pods controlled by application teams.

### DIFF
--- a/cmd/network-sidecar/cmd/run.go
+++ b/cmd/network-sidecar/cmd/run.go
@@ -77,16 +77,8 @@ func newCmdRun() *cobra.Command {
 			}
 			ctrl.SetLogger(ctrlzap.New(ctrlzap.UseFlagOptions(&zapOpts)))
 
-			// Create a basic zap logger for the log level server
-			zapConfig := zap.NewProductionConfig()
-			zapConfig.Level = atomicLevel
-			basicLogger, err := zapConfig.Build()
-			if err != nil {
-				return fmt.Errorf("failed to create logger for log level server: %w", err)
-			}
-
 			// Start dynamic log level server (non-blocking, non-fatal)
-			log.StartDynamicLevelServer(cfg.LogLevelAPI, atomicLevel, basicLogger)
+			log.StartDynamicLevelServer(cfg.LogLevelAPI, atomicLevel, ctrl.Log)
 
 			return nil
 		},
--- a/cmd/network-sidecar/cmd/run.go
+++ b/cmd/network-sidecar/cmd/run.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -29,12 +31,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
-	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	ctrlzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	meridio2v1alpha1 "github.com/nordix/meridio-2/api/v1alpha1"
 	"github.com/nordix/meridio-2/internal/common/config"
+	"github.com/nordix/meridio-2/internal/common/log"
 	"github.com/nordix/meridio-2/internal/controller/sidecar"
 )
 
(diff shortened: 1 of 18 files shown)


ljkiraly (approved):
Looks good.
The implementation is clean, well-tested, and follows the design specified in issue #196 

Clear improvement over signal-based toggle (SIGUSR1/2). HTTP gives arbitrary level selection, query-before-set, per-container control, and standard automation support.

**Add a one-line note about `--zap-*` flag removal**: The PR removes unused `--zap-*` flags inherited from kubebuilder scaffolding. These were never documented or referenced anywhere in the project, so this is a cleanup rather than a breaking change. A brief mention in the PR description is sufficient for traceability.

arp-est:
AI found the following:

```
  1. Missing documentation update for docs/controllers/loadbalancer.md

  The loadbalancer controller flags table at line 672 still shows:

  | `--log-level` | `MERIDIO_LOG_LEVEL` | `info` | Log level (`debug`, `info`, `error`) |
```
  
  
  Otherwise I think it looks good.

bszirtes:
> AI found the following:
> 
> ```
>   1. Missing documentation update for docs/controllers/loadbalancer.md
> 
>   The loadbalancer controller flags table at line 672 still shows:
> 
>   | `--log-level` | `MERIDIO_LOG_LEVEL` | `info` | Log level (`debug`, `info`, `error`) |
> ```
> 
> Otherwise I think it looks good.

Resolved in 4a51a56 .

arp-est (approved):

ljkiraly:
Each test that calls StartDynamicLevelServer leaks a goroutine and holds a  port until the test process exits. The tests use different ports (19901-19905) to avoid conflicts, but with go test -count=N or race detector, leaked goroutines accumulate. Not a bug today, but a test hygiene issue.
"nice-to-have": accept a context.Context, call server.Shutdown(ctx) on cancellation, return the server so tests can close it.

bszirtes:
> Each test that calls StartDynamicLevelServer leaks a goroutine and holds a port until the test process exits. The tests use different ports (19901-19905) to avoid conflicts, but with go test -count=N or race detector, leaked goroutines accumulate. Not a bug today, but a test hygiene issue. 
"nice-to-have": accept a context.Context, call server.Shutdown(ctx) on cancellation, return the server so tests can close it.

Resolved in [d9b722a](https://github.com/Nordix/Meridio-2/pull/199/commits/d9b722a6e92b96a70a8c2cb07ab280d504d4f0e6) .

Issue #115, opened 2026-04-24
Static+BFD: e2e tests


zolug:
Requires core Static+BFD functionality to be in place.

elajkat:
We had a chat with @arp-est and @bszirtes :

- for Static + BFD we need some kind of dual-stack suite
- the best would be to have a new static suite which can be tested both IPv4 and IPv6

arp-est:
For now we decided to only run the ipv4 appnetwork tests that are already present for bgp, also on a static setup.

I have no idea currently what extra benefit we would get from a dual-stack setup.
I mean that we have the ipv4 tests run on a static setup, that will check if the setup is capable of working, but setting a dual-stack suite I don't feel necessary since bgp already tests ipv6 setup, it should work the same on static, just with the need to manually setup the routes.
So the route setup is the only difference in theory, though I'm up for any suggestions.

PR #200, opened 2026-07-14
Add static bfd monitor, plus tests

arp-est:
https://github.com/Nordix/Meridio-2/issues/104
https://github.com/Nordix/Meridio-2/issues/115

AI summary:
```  Static+BFD Readiness Gates

  What it does

  Extends the connectivity monitoring to support static routing with BFD. Previously, readiness gates (meridio-2.nordix.org/ipv4-connectivity / ipv6-connectivity) only worked for
  BGP sessions. Now, for GatewayRouters using static routing with BFD configured, the readiness gate reflects the BFD session health — if BFD goes down, the protocol is marked as
  not established, the gate goes False, and the ENC controller stops routing traffic through that LB pod.

  How it works

  1. Monitor queries both birdc show protocols and birdc show bfd sessions each tick
  2. applyBfdState() in run.go mutates static protocol statuses:
    - Static + BFD configured + BFD session Up → Info = "Established"
    - Static + BFD configured + BFD session Down/missing → State = Down
    - Static + no BFD → Info = "Established" (always connected)
    - BGP → untouched (BIRD already reports correct state)

  3. All downstream logic (protocolsUp, ClassifyConnectivityByFamily, gate manager) just uses IsEstablished() uniformly — no protocol-specific branching needed

  Files modified

  - internal/bird/monitor.go — BfdSession struct, parseBfdOutput(), BfdSessions on MonitorStatus, BFD query in checkStatus(), BgpInfoEstablished constant
  - internal/controller/router/connectivity.go — unchanged (original simple ClassifyConnectivityByFamily)
  - cmd/router/cmd/run.go — applyBfdState(), routers fetched earlier in monitor loop, passed to mutation before downstream logic

  Tests

  - cmd/router/cmd/run_test.go — unit tests for applyBfdState (8 cases: BFD up/down/missing, no BFD, BGP untouched, mixed protocols, etc.)
  - api/v1alpha1/gatewayrouter_validation_test.go — added test confirming empty static: {} (no BFD) is accepted by the API

  E2E suite

  - test/e2e/suites/separate-static-appnetwork/ — new suite with static+BFD routing, VIPs 110.0.0.1/110.0.0.2, VLANs 1100/1200
  - VPN gateway configured with 10 static routes per VIP with BFD — allows scaling to 10 replicas without gateway reconfiguration
  - NADs use range_start/range_end to constrain LB pod IPs to .1-.10
  - Added to SUITES_IPV4 in test/e2e/Makefile and testCases in e2e_suite_appnetwork_test.go
  - Updated all e2e routing.yaml files to explicitly set protocol: "BGP" or "Static"
  - Updated VPN gateway (init.sh, bird-gw.conf) addresses from 169.254.60/61 to 169.254.110/111

  API fixes

  - Added +kubebuilder:validation:Format="" + Maximum=4294967295 + Minimum=0 on RemoteASN/LocalASN to allow 4-byte ASNs
  - Removed the immutability of BGPSpec, StaticSpec, and Protocol fields.
  ```

### DIFF
--- a/api/v1alpha1/gatewayrouter_types.go
+++ b/api/v1alpha1/gatewayrouter_types.go
@@ -39,7 +39,6 @@ type GatewayRouterSpec struct {
 
 	// protocol selects the routing protocol for this peering.
 	// +kubebuilder:validation:Enum=BGP;Static
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="protocol is immutable"
 	Protocol RoutingProtocol `json:"protocol"`
 
 	// bgp defines BGP session parameters. Required when protocol is BGP.
@@ -47,21 +46,35 @@ type GatewayRouterSpec struct {
 	// If the Protocol is bgp, the minimal parameters to be defined in bgp properties
 	// are RemoteASN and LocalASN
 	// +optional
-	// +kubebuilder:validation:XValidation:rule="!oldSelf.hasValue() || self == oldSelf.value()",message="bgp is immutable once set",optionalOldSelf=true
 	BGP *BgpSpec `json:"bgp,omitempty"`
 
 	// static defines static routing with BFD supervision. Required when protocol is Static.
 	// +optional
-	// +kubebuilder:validation:XValidation:rule="!oldSelf.hasValue() || self == oldSelf.value()",message="static is immutable once set",optionalOldSelf=true
 	Static *StaticSpec `json:"static,omitempty"`
 }
 
 type BgpSpec struct {
 	// The ASN number of the Gateway Router
+	//
+	// Note: Format="" suppresses the default int32 format that kubebuilder generates
+	// for uint32 fields. Without it, 4-byte ASNs (> 2147483647) would be rejected
+	// by the CRD schema validation. The explicit Maximum/Minimum enforce the full
+	// uint32 range at the API level.
+	// +kubebuilder:validation:Format=""
+	// +kubebuilder:validation:Maximum=4294967295
+	// +kubebuilder:validation:Minimum=0
 	// +required
 	RemoteASN uint32 `json:"remoteASN"`
 
 	// The ASN number of the system where the Attractor FrontEnds locates
+	//
+	// Note: Format="" suppresses the default int32 format that kubebuilder generates
+	// for uint32 fields. Without it, 4-byte ASNs (> 2147483647) would be rejected
+	// by the CRD schema validation. The explicit Maximum/Minimum enforce the full
+	// uint32 range at the API level.
+	// +kubebuilder:validation:Format=""
+	// +kubebuilder:validation:Maximum=4294967295
+	// +kubebuilder:validation:Minimum=0
 	// +required
 	LocalASN uint32 `json:"localASN"`
 
--- a/cmd/router/cmd/run.go
+++ b/cmd/router/cmd/run.go
@@ -21,6 +21,7 @@ import (
 	"crypto/tls"
 	"fmt"
 
+	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
 	"golang.org/x/sync/errgroup"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -244,33 +245,35 @@ func monitorConnectivity(
 				return fmt.Errorf("monitor channel closed unexpectedly")
 			}
 
+			// Fetch routers from informer cache (local memory read)
+			routers, err := getGatewayRoutersFromCache(ctx, mgr.GetClient(), cfg.GatewayName, cfg.GatewayNamespace)
+			if err != nil {
+				log.Error(err, "failed to get gateway routers from cache")
+				continue
+			}
+
+			// Mutate static protocol states based on BFD session status.
+			// After this, all protocols reflect true connectivity (Info != Established if BFD is down).
+			applyBfdState(status.Protocols, status.BfdSessions, routers, log)
+
 			count := protocolsUp(status.Protocols)
 
 			// Log when protocol up count changes
 			if firstUpdate || count != lastCount {
-				if status.HasConnectivity {
-					log.Info("Gateway connectivity established", "status", status.StatusString())
-				} else {
-					log.Info("Gateway connectivity lost", "status", status.StatusString())
-				}
+				log.Info("Gateway connectivity", "protocols", fmt.Sprintf("%d/%d up", count, len(routers)))
 				lastCount = count
 				firstUpdate = false
 			}
 
 			// Update readiness gates
 			if gateMgr != nil {
-				// Rebuild family map on every tick — reads from informer cache (no API call)
-				// and BuildFamilyMap is a trivial loop. Handles GatewayRouter add/remove/replace.
-				routers, err := getGatewayRoutersFromCache(ctx, mgr.GetClient(), cfg.GatewayName, cfg.GatewayNamespace)
-				if err == nil {
-					var ipv4, ipv6 bool
-					if len(routers) > 0 {
-						ipv4, ipv6 = router.ClassifyConnectivityByFamily(status.Protocols, router.BuildFamilyMap(routers))
-					}
-					// When no routers exist, ipv4/ipv6 stay false — gates set to False
-					if err := gateMgr.OnStatusUpdate(ctx, ipv4, ipv6); err != nil {
-						log.Error(err, "failed to update readiness gates")
-					}
+				var ipv4, ipv6 bool
+				if len(routers) > 0 {
+					ipv4, ipv6 = router.ClassifyConnectivityByFamily(status.Protocols, router.BuildFamilyMap(routers))
+				}
+				// When no routers exist, ipv4/ipv6 stay false — gates set to False
+				if err := gateMgr.OnStatusUpdate(ctx, ipv4, ipv6); err != nil {
+					log.Error(err, "failed to update readiness gates")
 				}
 			}
 		}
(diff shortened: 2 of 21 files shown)

### CODE REVIEW

**Review by bszirtes — COMMENTED** [review]
The e2e test part looks good.

File: cmd/router/cmd/run.go [hunk 1]
```diff
@@ -244,6 +244,17 @@ func monitorConnectivity(
...
 
+			// Fetch routers from informer cache (local memory read)
+			routers, err := getGatewayRoutersFromCache(ctx, mgr.GetClient(), cfg.GatewayName, cfg.GatewayNamespace)
+			if err != nil {
+				log.Error(err, "failed to get gateway routers from cache")
+				continue
+			}
+
+			// Mutate static protocol states based on BFD session status.
+			// After this, all protocols reflect true connectivity (State == Down if BFD is down).
```
ljkiraly: If I understand correctly the function sets the p.Info (not p.State) or which State is referred here?

internal/bird/monitor.go
```diff
@@ -157,6 +174,49 @@ func hasConnectivity(protocols []ProtocolStatus) bool {
...
+			continue
+		}
+
+		fields := strings.Fields(line)
+		if len(fields) < 3 {
+			continue
+		}
+
+		// Skip protocol name lines like "bfd1:"
+		if len(fields) == 1 && strings.HasSuffix(fields[0], ":") {
```
ljkiraly: This filter is never used. The previous one (in line 201) catch the protocol name also. 
len (["bfd1:"]) < 3  == true (1 field)
Remove the dead code.

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -58,10 +58,16 @@ type GatewayRouterSpec struct {
 
 type BgpSpec struct {
 	// The ASN number of the Gateway Router
+	// +kubebuilder:validation:Format=""
+	// +kubebuilder:validation:Maximum=4294967295
```
ljkiraly, line 64: Add some explanation to the comment. 
Validation is correct, allows full uint32 range for 4-byte ASNs.

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -58,10 +58,16 @@ type GatewayRouterSpec struct {
...
+	// +kubebuilder:validation:Minimum=0
 	// +required
 	RemoteASN uint32 `json:"remoteASN"`
 
 	// The ASN number of the system where the Attractor FrontEnds locates
+	// +kubebuilder:validation:Format=""
+	// +kubebuilder:validation:Maximum=4294967295
```
ljkiraly, line 76: Add some explanation to the comment.
Validation is correct, allows full uint32 range for 4-byte ASNs.

cmd/router/cmd/run.go
```diff
@@ -307,3 +313,45 @@ func protocolsUp(protocols []bird.ProtocolStatus) int {
...
+		gr := routerByName[p.Name]
+		if gr == nil {
+			continue
+		}
+
+		if gr.Spec.Static != nil && gr.Spec.Static.BFD != nil {
+			if bfdByIP[gr.Spec.Address] {
+				p.Info = bird.BgpInfoEstablished
+			} else {
+				p.Info = bird.BfdInfoDown
```
ljkiraly, line 353: A debug log when BFD causes a static protocol to be marked down helps production debugging.
```suggestion
				p.Info = bird.BfdInfoDown
				log.V(1).Info("BFD down for static protocol", "protocol", p.Name, "address", gr.Spec.Address)
```
Requires passing a `logr.Logger` into `applyBfdState` (add `log logr.Logger` parameter), or alternatively use `slog` package-level logging. 
Could be a follow-up.

ljkiraly (commented):
Excellent design and functionally good.
After minor fixes can be approved. 
Beside my online suggestions one test coverage gap:
- Consider to add unit tests for `parseBfdOutput` - at minimum test standard BIRD output format, empty output, and IPv6 sessions. `parseBfdOutput` only tested indirectly through `applyBfdState`. 
Should test:
   - Standard BIRD BFD output with multiple sessions
   - Empty output (no BFD configured in BIRD)
   - Output with only headers (BFD protocol exists but no sessions)
   - IPv6 BFD sessions
   - Malformed lines

cmd/router/cmd/run.go
arp-est: Ah, yes, originally the AI set state := Down, but later I've changed it to use the Info field instead, I didn't want it to mutate the state field itself.
I'll update the comment

cmd/router/cmd/run.go
arp-est: I'll pass the log as parameter.

internal/bird/monitor.go
arp-est: removed

api/v1alpha1/gatewayrouter_types.go
arp-est: done

api/v1alpha1/gatewayrouter_types.go
arp-est: done

cmd/router/cmd/run.go
zolug: HasConnectivity does not respect BFD status in case of Static, thus logging can be misleading. No need to pre-calculate it in checkStatus(), especially since count conveys the same meaning. Can be dropped from status alltogether.

cmd/router/cmd/run.go
```diff
@@ -307,3 +314,47 @@ func protocolsUp(protocols []bird.ProtocolStatus) int {
...
+		if p.Proto != "Static" || !p.State.IsUp() {
+			continue
+		}
+
+		gr := routerByName[p.Name]
+		if gr == nil {
+			continue
+		}
+
+		if gr.Spec.Static != nil && gr.Spec.Static.BFD != nil {
```
zolug, line 349: Would be nice to match against the interface as well. In meridio v1 there could be only 1 external interface per router container, but here there could be many at least in theory.

internal/bird/monitor.go
rubasov: Is this TODO resolved by this PR? If yes, please remove it.

cmd/router/cmd/run.go
arp-est: added the interface part

cmd/router/cmd/run.go
arp-est: removed hasConnectivity

arp-est:
> Excellent design and functionally good. After minor fixes can be approved. Beside my online suggestions one test coverage gap:
> 
>     * Consider to add unit tests for `parseBfdOutput` - at minimum test standard BIRD output format, empty output, and IPv6 sessions. `parseBfdOutput` only tested indirectly through `applyBfdState`.
>       Should test:
>       
>       * Standard BIRD BFD output with multiple sessions
>       * Empty output (no BFD configured in BIRD)
>       * Output with only headers (BFD protocol exists but no sessions)
>       * IPv6 BFD sessions
>       * Malformed lines

added tests

internal/bird/monitor.go
```diff
@@ -110,6 +118,15 @@ func (b *Bird) checkStatus(ctx context.Context) MonitorStatus {
 	}
 
 	status.Protocols = parseProtocolOutput(string(out))
+
+	bfdCmd := exec.CommandContext(ctx, "birdc", "-s", b.SocketPath, "show", "bfd", "sessions")
+	bfdOut, err := bfdCmd.CombinedOutput()
+	if err != nil {
+		log.V(1).Info("birdc bfd query failed", "error", err)
```
rubasov, line 122: It can be helpful in debugging if log lines like these always contain the command that failed, something like:

`Info("birdc failed", "cmd", bfdCmd, "error", err)`

Line 116 is not added by this PR, but this comment may apply there too.

arp-est:
Added a follow up issue to discuss the future of the "Gateway established", "-//-lost" printouts
https://github.com/Nordix/Meridio-2/issues/223

internal/bird/monitor.go
```diff
@@ -147,14 +160,42 @@ func parseProtocolOutput(output string) []ProtocolStatus {
...
+	State     string // "Up", "Down", "Init", etc.
+}
+
+// IsUp returns true if the BFD session state is "Up".
+func (b BfdSession) IsUp() bool {
+	return b.State == "Up"
+}
+
+// parseBfdOutput parses the output of `birdc show bfd sessions`.
+func parseBfdOutput(output string) []BfdSession {
```
rubasov, line 175: * This function could have been named parseShowBfdSessions.

* For maintainers coming back to this it can be helpful to add an example of the output we are parsing as a comment here.

* By the way I just found the '-v' flag for birdc in the bird docs:
https://bird.network.cz/doc/bird-4.html

Which is probably a feature to help parse the output of birdc. Here's an example I just made up, which leaves just a table to parse:

```
# birdcl -r -v -s /run/bird1/bird.ctl show protocols all | egrep '^1002-' | cut -d- -f2- 
kernel1    Kernel     master4    up     07:57:05.689  
static_routes Static     master4    up     07:57:05.689  
bird2      BGP        ---        up     08:28:11.415  Established   
device1    Device     ---        up     07:57:05.689  
direct1    Direct     ---        up     07:57:05.689  
```

rubasov (commented):
As a new contributor I started reviewing this to learn about the codebase and the PRs in flight. Feel free to use or ignore my comments as you see fit. They are not meant to block your PR in any way.

internal/bird/monitor.go
arp-est: yeah, thanks good catch

internal/bird/monitor.go
arp-est: I havent thought about it, previously we had only a single birdc command that could fail, I can add it.

internal/bird/monitor.go
arp-est: For point 1, and 3, Renaming it to the same as the command and adding the -v flag to have some return codes seems like a good idea, but then I would need to update the parseProtocolOutput function name too, for consistency.
How about we make a follow up issue about these two things?

For point 2: There are units tests that already have sample birdc output I don't feel the need of an extra comment.

ljkiraly (approved):
Looks good now

internal/bird/monitor.go
arp-est: oh wait, it seems to already distinguish the only two possible queries
`
log.V(1).Info("birdc query failed", "error", err) // show protocols

log.V(1).Info("birdc bfd query failed", "error", err) // show bfd sessions
   `
So I would prefer to leave it as is in this pr, we can add it to the follow up issue mentioned below

cmd/router/cmd/run.go
```diff
@@ -244,33 +245,41 @@ func monitorConnectivity(
...
+			applyBfdState(status.Protocols, status.BfdSessions, routers, log)
+
 			count := protocolsUp(status.Protocols)
 
 			// Log when protocol up count changes
 			if firstUpdate || count != lastCount {
-				if status.HasConnectivity {
-					log.Info("Gateway connectivity established", "status", status.StatusString())
+				if count == 0 {
+					log.Info("Gateway connectivity ALL DOWN", "status", status.StatusString())
```
zolug: Wording is weird. Connectivity can be interpreted (and was in case of Meridio v1) as sg above Gateway Connections.
"All Gateway Connections Down", "All Gateway Connection Up", "Some Gateway Connection Down" or similar at this point would make more sense.

Btw, what happens here during monitoring if there are no GatewayRouters by config (i.e. len(routers)==0)? Probably a dedicated log would be worth for that case. Does it make sense to call applyBfdState() if len(routers) is zero?

docs/controllers/router.md
```diff
@@ -124,13 +124,14 @@ Blackhole routes (`0.0.0.0/0` and `::/0`) in table 4097 prevent traffic leaking
...
+### BGP and BFD Monitoring
 
-A separate goroutine polls `birdc show protocols all "NBR-*"` at 1-second intervals:
+A separate goroutine polls `birdc show protocols all "NBR-*"` and `birdc show bfd sessions` at 1-second intervals:
 
 - Parses protocol state (up/down/start/idle) and BGP session info (Established)
-- Tracks connectivity: at least one `NBR-*` protocol in state `up` with info `Established`
-- Currently log-only; status changes logged when protocol up-count changes
+- For static protocols with BFD configured, applies BFD session state: if the BFD session is Up, Info is set to `Established`; if BFD is down or not found, Info is set to `BFD Down`
+- Tracks connectivity: at least one `NBR-*` protocol with Info containing `Established`
```
zolug: Is this what gets exposed via the readiness gates? If so, would be nice to explain connectivity is handled separately for IPv4 and IPv6

internal/bird/monitor.go
rubasov: Okay.

internal/bird/monitor.go
rubasov: I'm completely okay with a followup. After commenting here I also found the examples in the tests, thanks!

cmd/router/cmd/run.go
arp-est: I don't like the Some Gateway Connection Down message, also it would be nice for them to be uniform so that its easier to grep, and I couldn't come up with any neat name for It otherwise.

I think we talked about this too much, how about just print out one line with protocol count/total, and just get it done, and just do the discussion in the follow up.

docs/controllers/router.md
arp-est: discussed it, updated it to use your suggestion

Issue #172, opened 2026-06-15
Check controllers for non-deterministic ordering causing unnecessary updates


zolug:
## Problem

Non-deterministic ordering in controller outputs causes unnecessary API writes and config regeneration. When a reconciler compares desired state against existing state using DeepEqual, any ordering difference is treated as a change — triggering an Update/write, which fires a watch event, which triggers another reconcile, potentially looping until ordering stabilizes.

The same applies to generated configuration files (e.g., BIRD config): if the generated output differs only in ordering, the config is rewritten for no good reason.

Root causes:
- Map iteration order in Go is non-deterministic
- Pod/object list order from the informer cache is not guaranteed
- These produce slices (next-hops, connections, domains, endpoints, config blocks) whose element order varies across reconciles

## Action

Audit all controllers for slices/maps whose ordering can vary between reconciles and that are either compared (DeepEqual), written to the API, or rendered into configuration. Sorting should be applied at the output boundary — i.e., just before the comparison or write — not necessarily at every intermediate processing step.

Controllers to check:
- `internal/controller/distributiongroup/` — EndpointSlice construction, Pod ordering
- `internal/controller/gateway/` — Deployment spec construction (labels, env vars, volumes)
- `internal/controller/loadbalancer/` — flow configuration, VIP sets
- `internal/controller/endpointnetworkconfiguration/` — gateway connections, domains, next-hops
- `internal/controller/router/` — BIRD config generation (BGP neighbors, VIPs, protocol blocks)
- `internal/controller/sidecar/` — route/rule ordering, VIP assignment
- `internal/nfqlb/` — target/route operations

Pattern to look for:
```go
for _, v := range someMap {       // non-deterministic iteration
    result = append(result, ...)  // produces non-deterministic slice
}
// result used in DeepEqual, written to API, or rendered to config → unnecessary churn
```

PR #208, opened 2026-07-15
Defensive ordering sorts (#172)

ljkiraly:
Add defensive sorts at output boundaries to prevent non-deterministic ordering
Addresses #172: 
- Adds sorting at output boundaries where map iteration or cache list order could produce non-deterministic slices that propagate to API writes or config generation.
  Add defensive sorts in these locations:
   - distributiongroup/gateways.go:listReferencedGateways — sort returned gateways slice by namespace/name. Update: the sort in listReferencedGateways doesn't actually prevent any unnecessary update or reconcile — it's purely cosmetic (maybe slightly nicer log output). It doesn't fit the issue's scope)
   - router/controller.go:getVIPs — sort vips before returning
   - bird/bird.go:generateConfig — sort data.IPv4VIPs and data.IPv6VIPs after splitting
- tests verify that outputs remain stable across repeated calls with the same input, catching non-deterministic ordering from map iteration.
- Skip BIRD config write and reload when unchanged

### DIFF
--- a/internal/bird/bird.go
+++ b/internal/bird/bird.go
@@ -172,6 +172,12 @@ func (b *Bird) generateConfig(vips []string, routers []*meridio2v1alpha1.Gateway
 		}
 	}
 
+	// Sort VIPs for deterministic config output: input order depends on
+	// Gateway status addresses which are pre-sorted, but sort defensively
+	// to guarantee stable BIRD config regardless of caller ordering.
+	slices.Sort(data.IPv4VIPs)
+	slices.Sort(data.IPv6VIPs)
+
 	slices.SortFunc(routers, func(a, b *meridio2v1alpha1.GatewayRouter) int {
 		return cmp.Compare(a.Name, b.Name)
 	})
--- a/internal/bird/bird.go
+++ b/internal/bird/bird.go
@@ -54,9 +54,10 @@ type Config struct {
 
 type Bird struct {
 	Config
-	nl      abstractNetlink
-	running bool
-	mu      sync.Mutex
+	nl         abstractNetlink
+	running    bool
+	lastConfig string // last written config for skip-if-unchanged guard
+	mu         sync.Mutex
 }
 
 func New(cfg Config) (*Bird, error) {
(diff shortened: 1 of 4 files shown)


ljkiraly:
Audit conclusion for remaining controllers:
- sidecar/ — all map iterations are imperative netlink ops (VIP add/del, rule/route sync, table flush). Order doesn't affect correctness. No changes needed.
- loadbalancer/ — flow and target operations are imperative NFQLB calls. IPs already sorted before AddTarget. No changes needed.
- nfqlb/ — imperative route/rule/target management. No API writes from map iteration. No changes needed.
- endpointnetworkconfiguration/ — already has explicit sorting + existing determinism tests (TestResolveGatewayConnections_DeterministicOrdering, TestGetSLLBRNextHops_DeterministicOrdering). 
 
No changes needed.

ljkiraly:
The sorted outputs produce a stable config string, and the string comparison prevents unnecessary file writes and BIRD reloads. 

Memory cost is ~2-5 KB for the lastConfig string.

arp-est (approved):
I'm fine with this, AI also didn't find anything noteworthy.

internal/controller/router/controller.go
```diff
@@ -165,6 +166,12 @@ func getVIPs(gateway *gatewayapiv1.Gateway) []string {
 			seen[addr.Value] = struct{}{}
 		}
 	}
+
+	// Sort for deterministic BIRD config generation: Gateway status addresses
+	// are pre-sorted by the gateway controller, but sort defensively to avoid
+	// config rewrites if that invariant is ever broken.
+	slices.Sort(vips)
```
zolug, line 173: FYI: The Controller Manager already sorts them before writing to Gateway.Status.Addresses: https://github.com/Nordix/Meridio-2/blob/master/internal/controller/gateway/routes.go#L106

But adding the same sort on the reader side is cheap and thus a good defensive practice. So no problem.

internal/bird/bird.go
```diff
@@ -146,9 +153,21 @@ func (b *Bird) Configure(ctx context.Context, vips []string, routers []*meridio2
...
+	// Skip file write and BIRD reload if config is unchanged.
+	// Relies on generateConfig producing deterministic output (sorted VIPs and routers).
+	if conf == b.lastConfig {
+		return nil
+	}
+
+	if err := b.writeConfigString(conf); err != nil {
 		return err
 	}
+	b.lastConfig = conf
```
zolug: What if `configure` fails after this? The `lastConfig` is already set to the new config, and the config file on the disk reflects the same content. Yet, Bird is still running with the old config, while the next reconcile could return early due to `conf == b.lastConfig` without reconfiguring Bird.

ljkiraly: Right, the conf should be saved into 'lastConfig' after Bird configured successfully.

zolug (approved):

PR #212, opened 2026-07-16
TCP Authentication Option

rubasov:
- **TCP-AO: Add TCP Authentication Option to GatewayRouter BGP**
- **TCP-AO: add e2e test**


rubasov:
This is where I intend to continue Lajos' work from https://github.com/Nordix/Meridio-2/pull/114.
It would have been nicer to continue in the original pull request, but as I learned I would have needed contributor access to his personal Meridio-2 fork to update that PR.
At the moment this was a test that I have the proper access to github - I do now.
Of course I need to eliminate the DNM patch.

Compared to where Lajos left his PR I responded to the comments about MAC conversion.

rubasov:
So far I addressed the following comments from #114:

* MAC algorithm names: On the API we accept any algorithm name. Which we pass down to BIRD without mangling them. Removed the conversion alogether. Therefore we accept all algo names that are accepted by BIRD. If a bad algo name comes in, BIRD will reject the generated config.
* Conflict in the use of VLAN 800: I believe this was resolved by Lajos already. He moved over to using VLANs 900 and 1000 in tcp-ao e2e tests. However he forgot to update a few comments, which I now updated.
* gw-b1 was renamed to gw-t1, but there were remnants of the old name in comments. Cleared up.
* Reverted the value of setkey in BIRD config to what it was (off) before this change. Anyway this should not change any behaviors beause this config option only matters on FreeBSD.
* When we don't have a CurrentKeyId, we no longer set all keys as preferred in BIRD config.

rubasov:
I rebased this PR to the tip of master. My current TODO list for this PR:

* Key rotation:
  * BIRD itself is happy with rotating keys only using the preferred attribute (not using the deprecated attribute). IMO that means we do not need to expose the deprecated attribute in the manifest.
  * BIRD can and does rotate keys without re-establishing the bgp session, the `Since` field in `birdc show protocols` will allow us to even assert this in a test.
  * On the other hand we recently turned the `Protocol`, `BGP` and `Static` fields immutable. Immutability of the `BGP.Authentication` key of course prevents key rotation. Discussing this we realized we don't have good reasons to keep these fields immutable, so I'm going to turn them back to mutable.
* Tests:
  * add test(s) for key rotation
* Rebase to eliminate the DNM patch.

docs/controllers/router.md
```diff
@@ -124,6 +124,68 @@ Blackhole routes (`0.0.0.0/0` and `::/0`) in table 4097 prevent traffic leaking
...
+- `hmac sha1`
+- `hmac sha224`
+- `hmac sha256`
+- `hmac sha384`
+- `hmac sha512`
+- `cmac aes128`
+
+**Key rotation**: Multiple keys can be configured in the keychain. The `currentKeyId` and `nextKeyId` fields control which key is active for sending and which is advertised for rotation. Both peers must share the same key material for the corresponding key IDs.
+
+**BIRD config output**: The controller generates BIRD `authentication ao` blocks with key entries for each keychain item whose Secret is successfully resolved. If any Secret referenced by a GatewayRouter's keychain is missing or unreadable, the reconciler retains the current BIRD configuration unchanged (the failure is logged) and waits for the next reconcile trigger. This ensures authentication is never applied with incomplete keys.
```
zolug: The phrasing can be misleading (asked Attila as well to read and he interpreted the description that only GatewayRouters with unresolved dependencies will be excluded the rest of the config will get updated.)

test/e2e/suites/README.md
```diff
@@ -74,8 +74,10 @@ IPv6 achieves this with `fd00:cafe:{X}::` (external) vs `fd00:cafe:1{X}0::` (int
 | 600 | sctp-multihoming | sctp-gw2 | `169.254.31.0/24` | — | `169.111.30.0/24` | — | `30.0.0.2/32` | 64517 | 4200000000 |
 | 700 | ipv4-simple | gw-m1 | `169.254.40.0/24` | — | `169.111.40.0/24` | — | `40.0.0.1/32` | 64518 | 4200000000 |
 | 800 | pod-cache-label | gw-pcl | `169.254.50.0/24` | — | `169.111.50.0/24` | — | `50.0.0.1/32` | 64519 | 4200000000 |
+| 900 | tcp-ao | gw-t1 | `169.254.103.0/24` | — | `169.111.100.0/24` | — | `60.0.0.1/32` | 64520 | 4200000000 |
```
arp-est: minor thing, but it would be better to continue the numbering here too 169.254.50.0 - > 169.254.60.0 (instead of 169.254.103.0), same for the internal ips

docs/controllers/router.md
```diff
@@ -124,6 +124,68 @@ Blackhole routes (`0.0.0.0/0` and `::/0`) in table 4097 prevent traffic leaking
...
+  interface: "vlan-100"
+  address: "169.254.100.150"
+  protocol: "BGP"
+  bgp:
+    localASN: 64512
+    remoteASN: 4200000000
+    authentication:
+      keychain:
+        - sendId: 1
+          receiveId: 1
```
arp-est: AI found that this is not consistent with what is described in the yaml (recvId)

test/e2e/suites/README.md
rubasov: Done.

docs/controllers/router.md
rubasov: Done.

docs/controllers/router.md
rubasov: Hopefully resolved the misleading part and added some rationale. Let me know if it's good enough.

rubasov:
Another rebase.

Regarding key rotation:
* BIRD itself is happy to rotate keys using only the preferred attribute (and not using the deprecated attribute). Therefore we do not need to expose the deprecated attribute in the manifest.
* BIRD can and does rotate keys without re-establishing the bgp session, the `Since` field in `birdc show protocols` will allow us to even assert this in a test.
* The last rebase made GatewayRouter's `spec.bgp.authentication` mutable again, so that no longer prevents us patching `CurrentkeyId` to rotate keys.
* Tested key rotation manually and everything worked as expected.
* I was looking at adding an automated test (or extending the e2e test we already have) for key rotation. At first I hoped to create a config where the vpn-gateway side can stay unchanged and the test only needs to modify the GatewayRouter's spec. But I got confused by BIRD's key selection behavior. So I either need to clear up that confusion or make the test modify both sides' key config.

So the refreshed TODO:
* add test(s) for key rotation

Other than that I believe this is ready for review.

rubasov:
Added a test for key rotation - its limitations are explained in code comments. So we have nothing left on the TODO list. Please review. (And please be gentle with my noob go skills.)

internal/controller/router/controller.go
```diff
@@ -193,6 +204,49 @@ func (r *RouterReconciler) gatewayRouterEnqueue(_ context.Context, obj client.Ob
 	return []ctrl.Request{{NamespacedName: client.ObjectKey{Name: r.GatewayName, Namespace: r.GatewayNamespace}}}
 }
 
+// secretReferencedPredicate filters Secret events to only those referenced by
+// a GatewayRouter's keychain that belongs to this controller's gateway.
+// This prevents listing GatewayRouters on every unrelated Secret event.
+func (r *RouterReconciler) secretReferencedPredicate() predicate.Funcs {
+	matches := func(obj client.Object) bool {
+		list := &meridio2v1alpha1.GatewayRouterList{}
+		if err := r.List(context.Background(), list, client.InNamespace(obj.GetNamespace())); err != nil {
```
arp-est: 2. context.Background() in predicate — Using context.Background() inside a predicate bypasses the manager's context lifecycle. If the manager is shutting down, this List call could hang or panic. This is a minor
  risk but worth noting — controller-runtime predicates don't receive a context, so this is a known awkward pattern.

internal/controller/router/controller.go
```diff
@@ -110,7 +119,9 @@ func (r *RouterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 func (r *RouterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&gatewayapiv1.Gateway{}).
-		Watches(&meridio2v1alpha1.GatewayRouter{}, handler.EnqueueRequestsFromMapFunc(r.gatewayRouterEnqueue))
+		Watches(&meridio2v1alpha1.GatewayRouter{}, handler.EnqueueRequestsFromMapFunc(r.gatewayRouterEnqueue)).
+		Watches(&corev1.Secret{}, handler.EnqueueRequestsFromMapFunc(r.secretEnqueue),
+			bldr.WithPredicates(r.secretReferencedPredicate()))
```
arp-est: I got into a discussion with AI, I can't decide myself, so I'm just putting it here. I don't think its a problem for us, but worth noteing.

```
 1. secretReferencedPredicate does a List on every Secret event — The predicate calls r.List(context.Background(), list, client.InNamespace(...)) to list all GatewayRouters. This runs on every Secret
  create/update/delete event that passes through the informer cache (since the cache watches all Secrets once RBAC allows it). In namespaces with many Secrets (TLS, image pull, etc.), this could be expensive.

     Consider caching the set of referenced Secret names during reconcile (or at least documenting the trade-off). Alternatively, a simpler approach: skip the predicate entirely and just let secretEnqueue fire — the
  reconcile is cheap and idempotent, and the work queue deduplicates. The predicate adds complexity for marginal savings.
```

It got into a lengthy wall of text, I'll try to summarize, so it was complaining that in the predicate, on *every* secret change will trigger a lookup for "which secret belongs to a gatewayrouter".
It suggested to remove the predicate and just trigger a reconcile on a secret change, arguing that this way if lets say 50 secrets change in a cycle it will only trigger a single reconcile, also that way its less code.
But on the downside this way we'll reconcile on every secret change in the namespace, not just the related ones.

api/v1alpha1/gatewayrouter_types.go
```diff
@@ -142,6 +146,53 @@ type BfdSpec struct {
...
+
+	// NextKeyId specifies the key ID the peer should transition to (maps to BIRD's rnext id).
+	// +optional
+	NextKeyId *uint8 `json:"nextKeyId,omitempty"`
+}
+
+// TcpAoKeyChain defines a single TCP-AO key configuration.
+type TcpAoKeyChain struct {
+	// SendId is the Send_ID for this key (0-255).
+	SendId uint8 `json:"sendId"`
```
arp-est, line 175: 5. No maximum validation on SendId / RecvId — The doc comment says "0-255" but there's no +kubebuilder:validation:Maximum=255 marker. Since these are uint8 in Go, they naturally cap at 255 — but the CRD schema
  shows type: integer without constraints, so the API server would accept values > 255 that wrap at the Go layer. Adding Maximum=255 to the CRD marker would make the API self-documenting and prevent confusion.

You probably need the same thing I had to add to the localASN, and remoteASN

arp-est (commented):
Otherwise I think the stuff I mentioned are minor issues, and the pr overall looks good.

api/v1alpha1/gatewayrouter_types.go
rubasov: Good catch.

internal/controller/router/controller.go
rubasov: Oh, the joys of review by LLMs. :-)

Fun fact: the predicate was introduced by Lajos, because of an earlier review comment by an LLM.

The original review comment was this (comment number 5):
https://github.com/Nordix/Meridio-2/pull/114#issuecomment-4779108696

The diff is here:
https://github.com/Nordix/Meridio-2/compare/a60e7cf4b565b8f8b06df9d5ae8cdcd8097bc224..92defdabb76e578eec717ff69b428240ce2ec949#diff-5ef05b8a5047f996105eef791aa777241f16e96e95341b2e515e790bb0fe7605

However the way he (or an LLM) implemented it, it was not reducing the number of GatewayRouter listings, so the code comment in front of it was definitely wrong.

To me, the whole thing had the smell of premature optimization, but I'm really new to k8s controller development, so what do I know... To gain some ground, I also had a conversation with my LLM. And it likely told me what *I* wanted to hear. :-)

Anyway, my takeaway from that conversation was: GatewayRouter listings go through the informer cache, which means our queries are answered from local memory and because of that, the cost of having many GatewayRouter listings is likely irrelevant here. So I reverted to Lajos' earlier version of enqueue-ing Secrets without a predicate. Which is the simplest code we had so far. And to my understanding, it already should perform well, because it reads from the local informer cache.

By the way: If you are interested, I attached my conversation with the LLM:
[secret-watch-controversy.txt](https://github.com/user-attachments/files/30590729/secret-watch-controversy.txt)

internal/controller/router/controller.go
rubasov: AFAIU if we kept the predicate around, we could not do anything about this, because predicates do not take context arguments. Maybe if we made the predicate a closure, we could have the context?

Anyway, the other change removed the predicate altogether, so the problem is gone.

rubasov:
@arp-est: Thanks for the review.

arp-est (commented):
Looks good to me

zolug:
Docs must be updated as well, mainly the constraints-and-limitations and the router controller doc.

It would make sense describing how key rotation works as well (maybe under docs/operations), instead of requiring to figure it out based on tests. This key rotation desc. could be covered in a separate issue though.

rubasov:
> Docs must be updated as well, mainly the constraints-and-limitations and the router controller doc.
> 
> It would make sense describing how key rotation works as well (maybe under docs/operations), instead of requiring to figure it out based on tests. This key rotation desc. could be covered in a separate issue though.

Updated: constraints-and-limitations, router controller doc
Added: docs/operations/tcp-ao-key-rotation.md

zolug (approved):

Issue #220, opened 2026-07-22
Remove dead node-event handling code and RBAC


zolug:
### Background

Issue #101 investigated whether the DG controller should react to Node NotReady events. The conclusion is that independent node failure detection is an **intentional architectural constraint** — premature Maglev ID reallocation is worse than the detection delay, and applications control the trade-off via Pod `tolerationSeconds` (see #101 for full rationale).

The current code contains a placeholder stub for node event handling that was never implemented and should be removed.

### Dead code to remove

1. **`mapNodeToDistributionGroup`** in `internal/controller/distributiongroup/mappers.go` — returns `nil`, does nothing
2. **`if enableTopology` branch** in `SetupWithManager` (`internal/controller/distributiongroup/controller.go`) — registers the no-op mapper
3. **`--enable-topology-hints` flag** and `EnableTopologyHints` config field (`internal/common/config/manager.go`)
4. **`MERIDIO_ENABLE_TOPOLOGY_HINTS` env binding** in `BindEnv`
5. **`&corev1.Node{}` cache entry** gated behind the flag (`pkg/app/app.go`)

### RBAC to remove

In `config/rbac/manager-clusterrole.yaml`:
```yaml
# DistributionGroup controller: topology-aware hints (optional, requires --enable-topology-hints)
- apiGroups:
  - ""
 resources:
  - nodes
 verbs:
  - get
  - list
  - watch
```

This grants cluster-wide read access to all Node objects (capacity, addresses, conditions, labels) despite the code never using it. Removing it reduces the controller-manager's privilege set.

### Rationale

- The mapper is a stub — it was added speculatively and never implemented
- The RBAC is granted unconditionally even though the flag defaults to `false`
- Dead code with broad RBAC is misleading for security reviews and audits
- If a DG-level node failure policy is ever needed, it can be designed and implemented properly at that time with appropriate RBAC scoped to the actual requirements

### Acceptance criteria

- [ ] All listed code removed
- [ ] `nodes` RBAC rule removed from `manager-clusterrole.yaml`
- [ ] `go build ./...` passes
- [ ] `make lint` passes
- [ ] Existing tests pass (no test references `enableTopology`)

Issue #101, opened 2026-04-13
Investigate if custom endpointSlice controller requires node events


zolug:
Kubernetes Node failure → Pod readiness chain:

1. Node becomes unreachable → kubelet stops sending heartbeats
2. After node-monitor-grace-period (default 40s), node controller marks Node as NotReady
3. After pod-eviction-timeout (default 5m), node controller starts evicting Pods — but this is for deletion, not readiness
4. However, the kubelet on the failed node is what sets PodReady condition. If the node is down, the kubelet can't update Pod status. The Pod's PodReady condition stays as it was (True) because no one updates it.

So, when a node disappears:
- The Pod's PodReady condition remains True (stale — kubelet is dead)
- The Pod's status.conditions are not updated by the control plane based on Node readiness
- Only after the taint-based eviction kicks in (default 5 minutes with node.kubernetes.io/not-ready taint + tolerationSeconds: 300) does the Pod get deleted

What the DG controller does:
- isPodReady() checks PodReady condition and DeletionTimestamp
- The endpoint's Ready field is set from isPodReady(&pwip.pod)
- The controller watches Pods via Watches(&corev1.Pod{}, ...)

The gap: Between node failure and Pod eviction (up to ~5m40s with defaults), the DG controller sees the Pod as Running with PodReady=True because the stale status is never updated. The endpoint stays Ready=true in the EndpointSlice. The LB controller continues routing traffic to an unreachable target.

The detection delay in Meridio v1 is ~60 seconds (the EntryTimeout default). This is configurable via the NSP_ENTRY_TIMEOUT env var.

zolug:
## Investigation Outcome

After analysis, we've concluded that the lack of independent node failure detection is an **intentional architectural constraint**, not a gap to be fixed.

### Why we do NOT react to Node NotReady events

Reacting to Node NotReady by immediately removing endpoints and reallocating Maglev IDs is harmful in the general case:

1. **Node unreachable ≠ node dead.** A network partition between the control plane and a worker node triggers NotReady, but application Pods on that node may still be running and serving traffic (e.g., the data-plane path is intact while the control-plane path is broken).

[...]

PR #222, opened 2026-07-22
Remove Node topology hints feature scaffolding (#101)

zolug:
#220 

### Description:

The `--enable-topology-hints` flag and associated code was scaffolding for a potential Node-watching feature that was never implemented (the mapper was a no-op stub returning nil).

After investigation (#101), the conclusion is that Node events are not appropriate for this controller:
- Node unreachable ≠ node dead — a control-plane partition triggers NotReady, but the Pod may still serve traffic on the data-plane path
- Premature Maglev ID reallocation disrupts active connections
- Kubernetes already provides the mechanism via Pod tolerationSeconds

This PR removes the dead code and documents the architectural decision.

**Changes:**
- Remove --enable-topology-hints flag and env binding
- Remove mapNodeToDistributionGroup stub
- Remove Node RBAC from ClusterRole
- Remove conditional Node cache config
- Replace "Future Enhancement" with "Node Failure Detection" rationale in DG doc

### DIFF
--- a/config/rbac/manager-clusterrole.yaml
+++ b/config/rbac/manager-clusterrole.yaml
@@ -1,6 +1,6 @@
 ---
 # ClusterRole for cluster-scoped resources only.
-# GatewayClass and Nodes are cluster-scoped and must be accessible
+# GatewayClass is cluster-scoped and must be accessible
 # regardless of --namespace.
 #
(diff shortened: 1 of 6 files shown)


bszirtes (approved):
Looks good to me. Nice cleanup work.

ljkiraly (approved):

Issue #206, opened 2026-07-14
Investigate NFQLB fwmark-based packet accounting for no-flow and no-target drops


zolug:
## Context

Currently, when NFQLB receives a packet that matches no flow or the selected shared-memory LB instance has no active targets, the packet is silently dropped (`NF_DROP`). This makes it impossible to account for these drops via standard Linux tooling. See #153 for the broader observability discussion.

The metrics proposal in #153 defines `meridio_2_lb_drops_total` (labels: `gateway`, `reason=no_flow|no_target`) sourced from nftables counters fed by NFQLB's `--nolb_fwmark` and `--notargets_fwmark` options. Independent of whether metrics are ever scraped, accounting for these drops via nftables counters provides immediate troubleshooting value (`nft list counters`).

## Proposal

Configure NFQLB with dedicated fwmark values for the two drop reasons. Instead of `NF_DROP`, NFQLB sets the fwmark and returns `NF_ACCEPT`, re-injecting the packet into the netfilter framework. Downstream nftables rules match on the fwmark, increment a named counter, and drop.

## Constraints

- **Multi-DG**: The solution must work correctly when multiple DistributionGroups are configured on the same LB Pod. The fwmark values and nftables rules should be shared across all DGs.
- **Scalability**: The nftables rule count must not grow with the number of DGs, flows, or targets. A fixed set of rules (one per drop reason) matching on fwmark is preferred over per-DG rules.

### Part 1: NFQLB fwmark behavior

- Validate that `--nolb_fwmark` and `--notargets_fwmark` work as expected (packet re-injected with mark, not dropped)
- Identify appropriate fwmark values that don't conflict with target identifier marks used for policy routing (target marks are in the range `offset + maglev_id`)
- Confirm that packets returning from NFQUEUE with NF_ACCEPT + fwmark continue processing
- Test under load: verify no packet leakage (marked packets must not reach the application)

### Part 2: nftables rules for accounting and drop

- Design nftables rules in nftable that:
  - Match packets with the no-flow fwmark → increment named counter → drop
  - Match packets with the no-target fwmark → increment named counter → drop
- Determine rule placement: rules go after the NFQUEUE rule in the prerouting chain (packets continue from where they left off after NFQUEUE returns them)
- Validate counter increments correctly (send traffic to non-existent flow / to a DG with no targets)
- Determine when these rules should be created (~Lifecycle Management)

## Expected outcome

A short report confirming:
[...]

PR #226, opened 2026-07-29
NFQLB fwmark-based packet accounting for no-flow and no-target drops

arp-est:
https://github.com/Nordix/Meridio-2/issues/206


arp-est:
It's a draft, but the task is done, strictly speaking, please look at https://github.com/Nordix/Meridio-2/issues/206 for the details.

ljkiraly (approved):
The design is sound and the implementation is good in general. 

Some recommendations:
- Add unit test for `NoLBFwmark()` / `NoTargetsFwmark()` derivation (trivial but documents the contract)
- Consider naming the nftables counters using named counters instead of anonymous `&expr.Counter{}` for easier `nft list counters` output (but per #206 report, inline counters are preferred for programmatic access via Go library), so definitely for debate.

**Note:** The test mock factory `func(queueNum, queueTotal uint16) (nftablesManager, error)` doesn't pass fwmark params. The real `NewManager` now takes 4 params. This works because tests use `NftManagerFactory` which bypasses the real constructor (mocks don't need fwmarks). Not a bug, but the factory type and real constructor have diverged in signature.

arp-est:
> * Add unit test for `NoLBFwmark()` / `NoTargetsFwmark()` derivation (trivial but documents the contract)

Added the tests, 
I'm on the opinion, that we do not need the named counters, the current method is good enough, the 'nft list ruleset' command prints it anyway, and according to ai, its easier to get them this way via the go lib.

internal/nfqlb/nfqlb.go
```diff
@@ -65,6 +71,16 @@ func New(options ...Option) (*NFQueueLoadBalancer, error) {
 	}, nil
 }
 
+// NoLBFwmark returns the fwmark value used when no flow matches.
+func (nfqlb *NFQueueLoadBalancer) NoLBFwmark() int {
+	return nfqlb.startingOffset - 2
```
zolug: I don't like how the drop fwmark values are derived explicitly from startingOffset:
It silently repurposes the semantics of startingOffset. startingOffset is documented as "starting offset for the forwarding mark to avoid collisions with existing routing tables". Nothing about that name or its existing contract suggests "also reserve 2 slots directly below me" for drop accounting. It also means these can't be reconfigured without moving the whole target range.

**Suggestion:**
- nolbFwmark/notargetsFwmark as their own nfqlbConfig fields with their own defaults (e.g., DefaultNoLBFwmark = 4998, DefaultNoTargetsFwmark = 4999), and WithNoLBFwmark
/WithNoTargetsFwmark options, consistent with queue/qlength/maxTargets in this package.
- Replace the >= 3 guard with real cross-field validation: nolbFwmark != notargetsFwmark, both > 0, and neither colliding with the target range (>= startingOffset at minimum).
- While here, export the default consts (DefaultQueue, DefaultStartingOffset, etc.) so internal/common/config/loadbalancer.go can reference nfqlb.DefaultQueue instead of duplicating 
"0:3" as a separate literal — that duplication already exists for --nfqueue and will bite us again once these fwmarks get CLI/env exposure.

arp-est: Yeah, I'm not saying its nice, but it does make a lot of things less complicated.
How would you check that its outside the ending range of the forwarding marks?, Or would you just say, that it should be strictly less then the startingOffset?
Also its only meant for us to get some debug counters. I don't think users would like to manage extra parameters, and think about them to not collide with the existing range.

I'd much rather repurpose the startingOffset to have the first two values (startingOffset+1, and +2) for nolb, and notargets fwmark, and have the dynamic range start after that, then have extra parameters in the code.

docs/controllers/distributiongroup.md
```diff
@@ -156,8 +156,8 @@ one address in any of the Gateway's configured subnets.
 **Why this matters for the LoadBalancer controller:**
 
 The LoadBalancer controller uses **dynamic ID offsets per DistributionGroup** to differentiate target IP routes:
-- Each DistributionGroup gets a contiguous fwmark range of size `maxEndpoints`, allocated dynamically starting at offset 5000
-- Routes are created with fwmarks: `fwmark = offset + maglev_id`
+- Each DistributionGroup gets a contiguous fwmark range of size `maxEndpoints`, allocated dynamically starting at `fwmarkBase+2` (default 5002 with `--fwmark-base=5000`)
+- Routes are created with fwmarks: `fwmark = startingOffset + maglev_id` (where `startingOffset` is the DG's allocated base)
```
zolug: Sg like this would make it more clear:
```
Routes are created with fwmarks: `fwmark = dgOffset + maglev_id`, where dgOffset is the per-DG base allocated by getOffset() (first DG gets 5002, next gets 5002 + maxEndpoints, etc.)
```

internal/controller/loadbalancer/controller_test.go
zolug: Laci already flagged this NftManagerFactory signature divergence. NewManager() now takes 4 (queueNum, queueTotal, nolbFwmark, notargetsFwmark).
Consider updating the factory signature to include the fwmark params, or at minimum add a comment on the factory type explaining why it intentionally diverges.

internal/nfqlb/const.go
```diff
@@ -26,10 +26,17 @@ const (
...
+
+// DefaultFwmarkBase is the default base fwmark value.
+// Layout: +0=nolb, +1=notargets, +2..=NFQLB instance offsets.
+const DefaultFwmarkBase = 5000
 
+// DefaultQueue is the default NFQUEUE range used by NFQLB.
+const DefaultQueue = "0:3"
+
+const (
 	maglevMMultiplier = 100
```
zolug: merely consmetics, but it's an unexported const so could be sited with the rest of the unexported consts.
Also, would be nice to consolidate `maxOffset`. It's somewhat hidden considering the nfqlb pkg has a dedicated const.go file. Not sure what are the plans around it, if it shall be exposed as a parameter. But validation of config.fwmarkBase could at least check if its below maxOffset (and there's room for drop fwmarks).

docs/operations/constraints-and-limitations.md
```diff
@@ -44,9 +44,9 @@ Resource changes in `GatewayConfiguration.spec.verticalScaling` trigger Pod recr
...
... fwmarks and routing table IDs dynamically per DistributionGroup. Each DG gets a contiguous range of size `maxEndpoints` (from the DG's Maglev configuration, default 100). Ranges are allocated sequentially starting at `fwmarkBase+2` (default 5002 with `--fwmark-base=5000`) and packed without gaps — the allocator finds the first non-overlapping range based on actual `maxEndpoints` values of existing instances. The formula is `fwmark = startingOffset + endpoint_identifier`, where `startingOffset` is the DG's allocated base within the range (first DG gets `fwmarkBase+2`, next gets `fwmarkBase+2+maxEndpoints`, etc.) and `fwmark == tableID`. The first two values below the instance range (`fwmarkBase+0` and `fwmarkBase+1`) are reserved for drop-accounting (no-flow and no-targets marks used in nftables). This range must not overlap with other fwmark or routing table usage in the LB Pod's network namespace. At startup, the LB cleans up all stale policy routing rules with `mark >= fwmarkBase+2` from a previous instance. DG offset assignment is in-memory — different LB Pods may assign different offsets to the same DistributionGroup. This is acceptable as routing tables are local to each Pod.
```
zolug: would be nice to improve clarity what an instance actually is

internal/controller/loadbalancer/controller_test.go
```diff
@@ -89,11 +89,15 @@ func newTestLBEPS(dg *meridio2v1alpha1.DistributionGroup, endpoints []meridio2v1
 
 // mockNFQLB mocks the NFQueueLoadBalancer for testing.
 type mockNFQLB struct {
-	instances map[string]*mockNFQLBInstance
+	instances      map[string]*mockNFQLBInstance
+	startingOffset int
```
zolug: This is confusing.  The mock should mirror the real implementation's model. The real NFQueueLoadBalancer stores fwmarkBase and derives everything from it.

internal/nfqlb/nfqlb.go
```diff
@@ -53,6 +53,12 @@ func New(options ...Option) (*NFQueueLoadBalancer, error) {
 		opt(config)
 	}
 
+	// fwmarkBase must be >= 1: fwmark 0 is reserved (means "no mark").
+	// Layout: fwmarkBase+0=nolb, fwmarkBase+1=notargets, fwmarkBase+2..=instance offsets.
+	if config.fwmarkBase < 1 {
+		return nil, fmt.Errorf("fwmarkBase must be >= 1 (got %d): fwmark 0 is reserved", config.fwmarkBase)
```
zolug, line 59: the comment about maxOffset was ignored

zolug (approved):

PR #228, opened 2026-07-30
Fix gateway enqueue l34route

### LINKED ISSUE #224: LB controller: gatewayEnqueue doesn't cover DGs linked only via L34Route
## Summary

`gatewayEnqueue()` only enqueues DistributionGroups that have direct `spec.parentRefs` pointing to the Gateway. A DG connected to the Gateway **only** through an L34Route (no direct `parentRefs`) is not re-reconciled when Gateway status changes.

## Impact

`getGatewayVIPs()` reads `Gateway.status.addresses` to populate nftables VIP sets. If Gateway addresses change and a DG has no direct `parentRef`:
- The `gatewayEnqueue` mapper does not trigger reconciliation for that DG
- nftables VIP sets are not updated until an unrelated event (e.g., EndpointSlice change, L34Route edit) triggers reconciliation
- New VIPs are not queued to nfqueue; removed VIPs continue to be queued unnecessarily

## Current Code

```go
// controller.go:gatewayEnqueue()
dgList := &meridio2v1alpha1.DistributionGroupList{}
c.List(ctx, dgList)

for _, dg := range dgList.Items {
    for _, parentRef := range dg.Spec.ParentRefs {
        if parentRef.Name == c.GatewayName && namespace == c.GatewayNamespace {
            requests = append(requests, ...)
        }
    }
}
```

Only checks `dg.Spec.ParentRefs` — never consults L34Routes.

## Proposed Fix

Extend `gatewayEnqueue` to also discover DGs via L34Routes:
1. List L34Routes with `parentRefs` matching this Gateway
2. Extract DG names from `backendRefs`
3. Enqueue those DGs (deduplicated with the direct parentRef set)

Alternatively, document as a requirement that DGs must have direct `parentRefs` when VIP updates need to propagate promptly.

## References

- PR #204 review comments (zolug, bszirtes)
- `internal/controller/loadbalancer/controller.go:gatewayEnqueue()`
- `docs/controllers/loadbalancer.md` — Watch Triggers section documents this gap

ljkiraly:
Extend gatewayEnqueue to discover DistributionGroups connected to the gateway both directly (via spec.parentRefs) and indirectly (via L34Route backendRefs). Previously, only DGs with direct parentRefs were enqueued when Gateway status changed.

Without this fix, a DG linked to a Gateway only through an L34Route would not be re-reconciled when Gateway.status.addresses changed, leaving nftables VIP sets stale until an unrelated event triggered
reconciliation.

Changes:
- gatewayEnqueue now scans L34Routes for backendRefs to DGs
- Deduplicates DGs found via both paths (map-based)
- Extract "Service" literal to defaultBackendKind constant (goconst)
- Add 5 unit tests covering direct, indirect, dedup, wrong-gateway,  and different-gateway-name cases
- Extract shared helper for parsing Gateway API BackendRef into a DistributionGroup ObjectKey. Replaces 3 duplicate instances of the same group/kind/namespace defaulting logic in:
     - belongsToGateway()
     - gatewayEnqueue()
     - l34RouteEnqueue()

Fixes: #224

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,6 +61,9 @@ Always use `kubebuilder create api` and `kubebuilder create webhook` to scaffold
 The e2e tests are designed to validate the solution in an isolated environment (similar to GitHub Actions CI).
 Ensure you run them against a dedicated [Kind](https://kind.sigs.k8s.io/) cluster (not your “real” dev/prod cluster).
 
+### Branch Naming
+Do not use `/` in branch names (e.g., `fix/my-feature`). The CI pipeline uses branch names as Docker image tags, and `/` is not valid in container image tags. Use hyphens instead: `fix-my-feature`.
+
 ## After Making Changes
 
 **After editing `*_types.go` or markers:**
--- a/docs/controllers/loadbalancer.md
+++ b/docs/controllers/loadbalancer.md
@@ -223,13 +223,13 @@ nftables (shared across all DGs in this LB Pod)
 |----------|---------|-----------------|-------------------|
 | DistributionGroup | Create/Update/Delete | Direct (`.For()`) | None (all events) |
 | LoadBalancerEndpointSlice | Create/Update/Delete | `endpointSliceEnqueue` | OwnerReference to DistributionGroup + `spec.gatewayRef` matches this Gateway |
-| Gateway | Create/Update/Delete | `gatewayEnqueue` | Name matches controller's Gateway + lists DGs with direct `parentRefs` to this Gateway |
+| Gateway | Create/Update/Delete | `gatewayEnqueue` | Name matches controller's Gateway + discovers DGs via direct `parentRefs` and indirect L34Route `backendRefs` |
 | L34Route | Create/Update/Delete | `l34RouteEnqueue` | Checks parentRefs for this Gateway AND backendRefs for DG kind |
 
 ### Filtering Strategy Rationale
 
 - **LoadBalancerEndpointSlice mapper**: Checks namespace matches this Gateway's namespace, `spec.gatewayRef` matches this Gateway, and ownerReference points to a DistributionGroup. Enqueues the owning DG.
-- **Gateway mapper**: Only triggers if Gateway name matches this controller's Gateway. Lists all DistributionGroups and enqueues those with `spec.parentRefs` referencing this Gateway. Note: DGs linked only via L34Route (no direct `parentRefs`) are not re-reconciled by this mapper — they rely on the L34Route mapper instead.
+- **Gateway mapper**: Only triggers if Gateway name matches this controller's Gateway. Discovers DGs via both direct `spec.parentRefs` and indirectly via L34Route `backendRefs`. Deduplicates results.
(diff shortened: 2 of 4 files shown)


arp-est (approved):
AI only found 1 nitpick worth mentioning

```

  1. c.List(ctx, dgList) — no namespace scoping. The DG list in the direct-parentRef path has no client.InNamespace(...). This is fine because the controller's cache is already scoped to GatewayNamespace (via
  cache.Options.DefaultNamespaces in run.go), so the list is implicitly namespace-filtered. But it's worth noting the asymmetry with the L34Route list which explicitly passes client.InNamespace.

> elaborate concern 1

  The concern is about this asymmetry in gatewayEnqueue:

  // Direct path — no namespace filter on List:
  dgList := &meridio2v1alpha1.DistributionGroupList{}
  if err := c.List(ctx, dgList); err != nil {

  // Indirect path — explicit namespace filter:
  l34routeList := &meridio2v1alpha1.L34RouteList{}
  if err := c.List(ctx, l34routeList, client.InNamespace(c.GatewayNamespace)); err != nil {

  Why it's fine: The LB controller's manager cache is configured with cache.Options.DefaultNamespaces scoped to GatewayNamespace in cmd/stateless-load-balancer/cmd/run.go. So c.List(ctx, dgList) without an explicit
  namespace filter still only returns objects from that namespace — the cache simply doesn't have objects from other namespaces.

  Why I flagged it: It's a readability/defensiveness concern, not a correctness bug. Someone reading gatewayEnqueue in isolation sees one list with explicit namespace scoping and one without, and might wonder if the
  unscoped list could return cross-namespace DGs. The answer is "no, the cache prevents it" — but that knowledge requires understanding the manager setup in a different file.

  Adding client.InNamespace(c.GatewayNamespace) to the DG list would make the code self-documenting and harmless (it's a no-op filter on an already-scoped cache). But it's genuinely optional — the current code is
  correct.
```

bszirtes (approved):
LGTM. The fix correctly closes the gap where DistributionGroups connected to a Gateway only via L34Route weren't re-reconciled on Gateway status changes, the `resolveBackendRefDG` extraction cleanly removes the previously triplicated BackendRef-defaulting logic, and the new tests cover the direct, indirect, dedup, and mismatch cases well; verified build, vet, gofmt, and all 42 unit tests pass locally. One minor, non-blocking note: `gatewayEnqueue`'s direct-parentRef path doesn't re-check `parentRef.Group` the way `belongsToGateway` does further down (it only compares `Name`/`Namespace`); since `Group` has no CEL validation guaranteeing it (unlike `Kind`, which the API server always enforces to `"Gateway"`), a `parentRef` with an explicitly overridden `Group` could still trigger an enqueue here, though in practice it only causes one extra no-op reconcile since `Reconcile()` re-validates membership via `belongsToGateway` anyway.

zolug:
I have created a follow-up issue: https://github.com/Nordix/Meridio-2/issues/230
Otherwise, it looks good to me.

zolug (approved):

PR #231, opened 2026-08-04
LB controller: minor cleanups (#225)

### LINKED ISSUE #225: LB controller: minor cleanups (dead code, no-op callback, rule priority)
## Summary

Three small independent cleanups identified during PR #204 review. All are low-risk, low-priority.

## Checklist

- [x] **Remove dead nftables variables from Controller struct and `run.go`**
  `runLoadBalancer()` creates a second nftables table (`"meridio"`) and passes `NFTConn`, `NFTTable`, `NFTChain` to the Controller — but these fields are never used. The real table (`"meridio-lb"`) is created inside `Controller.SetupWithManager()` via `NftManagerFactory`.
  Files: `cmd/stateless-load-balancer/cmd/run.go`, `internal/controller/loadbalancer/controller.go`
  ~20 lines removed.

- [x] **Remove no-op `updateNfQueueDestinationCIDRsFunc`**
  `AddFlow` and `DeleteFlow` call `s.updateNfQueueDestinationCIDRsFunc(ctx)` which is a permanent no-op (legacy from POC copy where nfqlb managed its own nftables VIP sets — now handled externally by `internal/nftables.Manager`).
  Files: `internal/nfqlb/nfqlb.go`
  ~10 lines removed.

- [x] **Set fixed priority on fwmark policy routing rules**
  `ensureRule` in `internal/nfqlb/routing.go` creates ip rules without an explicit priority (uses `netlink.NewRule()` default, which causes the kernel to auto-assign). This makes the `ensureRule` duplicate-detection unreliable (cannot match on priority when unset) and creates non-deterministic rule ordering. Should use a fixed priority constant (e.g., 32000).
  Files: `internal/nfqlb/routing.go`
  ~5 lines changed.

## References

- PR #204 review comments (zolug, bszirtes)
- Dead code: `cmd/stateless-load-balancer/cmd/run.go` lines 119-140
- No-op callback: `internal/nfqlb/nfqlb.go` AddFlow/DeleteFlow
- Rule priority: `internal/nfqlb/routing.go` getRule()

rubasov:
cleanup: Do not create nft table we never use after

Other than the meridio-lb nft table for some historical reason we also
created and deleted a table named meridio, which we never used after.
This patch deletes all code handling the meridio table.

Closes #225
Signed-off-by: Bence Romsics <bence.romsics@gmail.com>

cleanup: Remove no-op updateNfQueueDestinationCIDRsFunc

AddFlow and DeleteFlow called s.updateNfQueueDestinationCIDRsFunc(ctx)
which was legacy from the POC where nfqlb managed its own nftables
VIP sets.  Now internal/nftables.Manager handles this externally.
Therefore updateNfQueueDestinationCIDRsFunc became a permanent no-op.
This change removes updateNfQueueDestinationCIDRsFunc.

Related #225
Signed-off-by: Bence Romsics <bence.romsics@gmail.com>

cleanup: Set fix ip rule priority

ensureRule in internal/nfqlb/routing.go created ip rules without an
explicit priority (uses netlink.NewRule() default, which causes the
kernel to auto-assign). This makes the ensureRule duplicate-detection
less reliable and creates non-deterministic rule ordering. Start using
a fixed priority constant: 32000.

Related #225
Signed-off-by: Bence Romsics <bence.romsics@gmail.com>

### DIFF
--- a/internal/nfqlb/routing.go
+++ b/internal/nfqlb/routing.go
@@ -25,6 +25,8 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
+const rulePriority = 32000
+
 var errInvalidIP = errors.New("the ip address is invalid")
 
 // CleanupStaleRules removes all fwmark-based policy rules and their routing tables
@@ -73,9 +75,7 @@ func createPolicyRoute(fwMark int, ip string) error {
 
 // ensureRule adds the desired rule only if it doesn't already exist.
 // Linux allows duplicate ip rules, so we check first to avoid accumulating duplicates
-// across reconciles. Priority of -1 means "unset" in vishvananda/netlink (library skips
-// FRA_PRIORITY in the netlink message, causing the kernel to auto-assign a priority).
-// We cannot match on auto-assigned priorities, so skip the comparison when unset.
+// across reconciles.
 func ensureRule(desired *netlink.Rule) error {
 	rules, err := netlink.RuleList(desired.Family)
 	if err != nil {
@@ -85,7 +85,7 @@ func ensureRule(desired *netlink.Rule) error {
 
 	for _, existing := range rules {
 		if existing.Mark == desired.Mark && existing.Table == desired.Table &&
-			(desired.Priority < 0 || existing.Priority == desired.Priority) {
+			existing.Priority == desired.Priority {
 			return nil // Already exists
 		}
 	}
@@ -123,6 +123,7 @@ func getRoute(tableID int, ip net.IP) *netlink.Route {
 
 func getRule(fwMark int, ip net.IP) *netlink.Rule {
 	rule := netlink.NewRule()
+	rule.Priority = rulePriority
 	rule.Table = fwMark
 	rule.Mark = uint32(fwMark)
 	rule.Family = netlink.FAMILY_V6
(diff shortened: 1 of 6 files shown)


bszirtes (approved):
LGTM. All three cleanups are complete with no dangling references (`NFTConn`/`NFTTable`/`NFTChain`, `updateNfQueueDestinationCIDRsFunc`). Nice catch also updating `troubleshooting.md` to drop the stale `meridio` table example output.

One minor note for a future follow-up: `routing.go` (including the new fixed-priority `ensureRule` logic) has no dedicated unit tests. This is a pre-existing gap, not a blocker here.

arp-est (approved):
Looks good to me.

rubasov:
Thanks for the reviews! This is not a risky change, so I'll merge this now, not to leave it hanging for a week, while I'm out of office.

internal/nfqlb/routing.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
+const rulePriority = 32000
```
zolug, line 28: If there's a const.go file in the package please put new constants there (unless related to testing) to keep them centralized and avoid ambiguity.

rubasov: Thanks for the comment, I believe Botond already fixed this while I was out of office (thanks for that!).

Issue #223, opened 2026-07-23
Extend gateway connectivity log message indication + additional improvements :)


arp-est:
In cmd/router/cmd/run.go
There are a couple of log messages indication 1/2, 0/2, "Gateway connectivity established", or "lost".

It would be nice to extend it, so that it can show the following differences:

- [x] distinguish ipv4, or ipv6 protocols.
- [x] When let's say the output is 2/4, and then 2/4 again, then we want to check if the 2 protocols that are established are the same or different than on the previous log message.
In a bit more detail check for (A^,B^,C,D) vs (A,B^,C^,D) like changes, where ^ means that its established, and A,B,C,D are the protocol names here.

Additional improvements to router:
- [x] optimize applyBfdState() to early return in case of empty router list.
- [x] Rename parseProtocolOutput and parseBfdOutput to reflect the birdc commands executed.
- [x] Investigate birdc -v switch, it seems to be adding exit_codes to the birdc commands, and error messages based on the description could be useful, for executing the command via a program rather than by hand.
` Option -v can be passed to the client, to make it dump numeric return codes along with the messages. You do not necessarily need to use birdc to talk to BIRD, your own applications could do that, too -- the format of communication between BIRD and birdc is stable (see the programmer's documentation). ` 
- [x] Print out the birdc command when the execution fails.

Follow up for https://github.com/Nordix/Meridio-2/pull/200

arp-est:
Checked the output, of birdc, the -v switch seems to be a bit different than what I've imagined.
```
~ $ birdc -s /var/run/bird/bird.ctl 'show protocols' && echo "OK" || echo "FAIL"
BIRD 3.2.2 ready.
Name       Proto      Table      State  Since         Info
VIP4       Static     master4    up     07:10:38.276
NBR-gw-t1-router-v4 BGP        ---        start  07:10:38.276  Connect       Socket: Host is unreachable
device1    Device     ---        up     07:10:38.050
kernel1    Kernel     master4    up     07:10:38.050
kernel2    Kernel     master6    up     07:10:38.050
bfd1       BFD        ---        up     07:10:38.050
OK
~ $ birdc -vs /var/run/bird/bird.ctl 'show protocols' && echo "OK" || echo "FAIL"
0001 BIRD 3.2.2 ready.
2002-Name       Proto      Table      State  Since         Info
1002-VIP4       Static     master4    up     07:10:38.276
[...]

PR #233, opened 2026-08-07
Extend gateway connectivity log message indication + additional improvements :)

arp-est:
- Rename parseProtocolOutput and parseBfdOutput to reflect the birdc commands executed.
- Print out the birdc command when the execution fails.
- applyBfdState() early returns in case of empty router list.
- Change the status message whan a protocol status changes to be more informative. Now it print name, ipfamily, and both status/BGP status

https://github.com/Nordix/Meridio-2/issues/223


bszirtes (commented):
Good work overall: the renames correctly reflect the actual `birdc` commands used, `applyBfdState()`'s early return is correct, the command is now printed on `birdc` failure, and the `-v` switch investigation/conclusion is sound (it only prepends per-line reply codes, doesn't affect `birdc`'s exit code, so not worth adopting here).

A few things worth polishing before merge:

1. **`internal/bird/monitor.go:113,122`** — the error log call doesn't follow structured logging convention:

   ```go
   log.V(1).Info("'"+cmd.String()+"'", "failed, error: ", err)
   ```

   The message should be static, not the variable command string, and `"failed, error: "` isn't a valid key. Suggest:

   ```go
   log.V(1).Info("birdc command failed", "cmd", cmd.String(), "error", err)
   ```

2. **The connectivity log doesn't show the delta the issue asked for.** Right now, whenever anything changes, it reprints the *full* state of every protocol rather than highlighting which one(s) actually flipped — so two consecutive log lines still have to be diffed by hand to spot the `(A^,B^,C,D)` → `(A,B^,C^,D)`-style change. It's a solid improvement over the old `N/M up` counter, but stops short of surfacing the delta itself.
Two possible options: only log the protocols whose state actually changed, or keep the full snapshot but add an explicit changed field listing the names that flipped.

3. **No unit tests for the new helpers.** `protocolsEqual`, `mapsEqual`, and `protocolStateLabel` are all new in this diff (replacing `protocolsUp`) and drive the core logging logic, so they'd benefit from table tests covering nil vs empty slices/maps, reordering, and partial overlap.

rubasov:
I agree, this is a good change. Adds useful info to the logs that otherwise would have been lost forever. It also helps to read the code.

I fully agree with Botond's comment on having a static error message on birdc failure and then adding the command as an additional log field.

I'm not sure I agree with the suggestion of only logging the status diffs. To me the log is already easy to skim both visually and programmatically. This is simple to process even with many neighbors and many log lines:

    $ kubectl logs -n e2e-separate-static-appnetwork sllbr-gw-a1-6446cf8c58-bxhp7 -c router | egrep 'Gateway connectivity'
    {"level":"info","ts":"2026-08-08T11:50:19Z","logger":"monitor","msg":"Gateway connectivity","NBR-gw-a1-router-v4-static":"IPv4:up/Established"}

I think it has value if the full state is known at a certain time without the user needing to reconstruct it from the diffs in the log history. If we want to log the diffs too, I would lean to choosing Botond's 2nd suggestion: "add an explicit changed field listing the names that flipped."

(Btw, next week just go ahead, don't wait on my review, since I'll be on vacation.)

arp-est:
>     2. **The connectivity log doesn't show the delta the issue asked for.** Right now, whenever anything changes, it reprints the _full_ state of every protocol rather than highlighting which one(s) actually flipped — so two consecutive log lines still have to be diffed by hand to spot the `(A^,B^,C,D)` → `(A,B^,C^,D)`-style change. It's a solid improvement over the old `N/M up` counter, but stops short of surfacing the delta itself.
>        Two possible options: only log the protocols whose state actually changed, or keep the full snapshot but add an explicit changed field listing the names that flipped.


AI might have interpreted the task a bit strictly, I'm on the opinion that its a lot better to print the whole state then only the difference, as when debugging, it could be a lot of hassle to track down the previous state from the log history, also previous state changes could have already been rotated out.
I'll update the other two points though.

bszirtes (approved):
Points 1 and 3 look good, the log call now uses a static message with proper `cmd`/`error` keys, and the new table tests for `protocolStateLabel`, `protocolsEqual`, and `mapsEqual` cover the relevant edge cases well. On point 2, I take your point that printing the full snapshot is safer for debugging when log history gets rotated or a line is viewed out of context, that's a fair tradeoff and not something I'd block on; that said, full state and delta-visibility aren't mutually exclusive, you could keep the full snapshot as-is and just append an explicit `changed` field listing the names that flipped since the last log line, giving you both the complete picture and an at-a-glance answer to "what changed" without diffing two lines by hand, but this is just a suggestion, not a blocker.

internal/bird/monitor.go
```diff
@@ -110,25 +110,26 @@ func (b *Bird) checkStatus(ctx context.Context) MonitorStatus {
...
-	bfdCmd := exec.CommandContext(ctx, "birdc", "-s", b.SocketPath, "show", "bfd", "sessions")
-	bfdOut, err := bfdCmd.CombinedOutput()
+	cmd = exec.CommandContext(ctx, "birdc", "-s", b.SocketPath, "show", "bfd", "sessions")
+	out, err = cmd.CombinedOutput()
 	if err != nil {
-		log.V(1).Info("birdc bfd query failed", "error", err)
-	} else {
-		status.BfdSessions = parseBfdOutput(string(bfdOut))
+		log.V(1).Info("birdc command failed", "cmd", cmd.String(), "error", err)
+		return status
```
zolug, line 123: Would it make sense to introduce a `CheckError` or similar field to the MonitorStatus struct so that consumers of status reports could be informed that a status report cannot be fully trusted?
Currently, the monitor will go with the flow, and on momentary birdc hiccups (if that's even a thing) might come to the conclusion that everything is down.
Meridio-1 had the session error and toleration concept, to tolerate up to e.g. 3 consequtive status fetch errors before potentially downgrading external connectivity.
This would keep the checkStatus() method dumb, and monitorConnectivity() would be responsible for any policy decisions (e.g. to tolerate i.e. ignore 3 hiccups in a row).

**For example:**

```go
// MonitorStatus represents the overall monitoring status
type MonitorStatus struct {
    Protocols   []ProtocolStatus
    BfdSessions []BfdSession
    CheckError  error // non-nil if birdc queries failed this tick
}
```

Then in checkStatus():

```go
func (b *Bird) checkStatus(ctx context.Context) MonitorStatus {
    status := MonitorStatus{
        Protocols:   []ProtocolStatus{},
        BfdSessions: []BfdSession{},
    }

    // ... running check ...

    cmd := exec.CommandContext(ctx, "birdc", "-s", b.SocketPath, "show", "protocols", "all", `"NBR-*"`)
    out, err := cmd.CombinedOutput()
    if err != nil {
        status.CheckError = fmt.Errorf("%s: %w: %s", cmd.String(), err, strings.TrimSpace(string(out)))
        return status
    }
    status.Protocols = parseShowProtocolsOutput(string(out))

    cmd = exec.CommandContext(ctx, "birdc", "-s", b.SocketPath, "show", "bfd", "sessions")
    out, err = cmd.CombinedOutput()
    if err != nil {
        // BFD query failed but protocols are valid — partial success
        status.CheckError = fmt.Errorf("%s: %w: %s", cmd.String(), err, strings.TrimSpace(string(out)))
        // Don't clear Protocols — they're still valid
        return status
    }
    status.BfdSessions = parseShowBfdSessionsOutput(string(out))

    return status
}
```

And in monitorConnectivity():

```go
var sessionErrors int
const maxSessionErrors = 3  // or from config

for {
    select {
    case status, ok := <-statusCh:
        if !ok { ... }

        if status.CheckError != nil {
            sessionErrors++
            logger.V(1).Info("birdc check failed", "error", status.CheckError, "consecutive", sessionErrors)
            if sessionErrors <= maxSessionErrors {
                continue // preserve previous state, skip this tick
            }
            // Threshold exceeded
            logger.Info("birdc check threshold exceeded", "consecutive", sessionErrors)
            // Fall through with whatever partial data is in status
            // (empty protocols → gates go False)
        } else {
            sessionErrors = 0
        }

        // ... rest of the logic (applyBfdState, gates, logging) ...
    }
}

```

cmd/router/cmd/run.go
```diff
@@ -276,20 +277,36 @@ func monitorConnectivity(
...
-			count := protocolsUp(status.Protocols)
-
-			// Log when protocol up count changes
-			if firstUpdate || count != lastCount {
-				logger.Info("Gateway connectivity", "protocols", fmt.Sprintf("%d/%d up", count, len(routers)))
-				lastCount = count
+			familyMap := router.BuildFamilyMap(routers)
+
+			// Log when protocols or family map change
+			if firstUpdate || !protocolsEqual(status.Protocols, lastProtocols) || !mapsEqual(familyMap, lastFamilyMap) {
```
zolug: Why not use maps.Equal(a, b) instead?

cmd/router/cmd/run.go
```diff
@@ -276,20 +277,36 @@ func monitorConnectivity(
...
+				birdProtocols := make(map[string]bool, len(status.Protocols))
+				kvs := make([]any, 0, len(familyMap)*2)
+				for _, p := range status.Protocols {
+					if family, ok := familyMap[p.Name]; ok {
+						kvs = append(kvs, p.Name, family+":"+protocolStateLabel(p))
+						birdProtocols[p.Name] = true
+					}
+				}
+				// Log routers configured but not yet in BIRD
+				for name, family := range familyMap {
```
zolug: It could help diffing (especially "not yet configured in bird") log entries if familyMap was sorted by keys before this (its a map)

zolug:
Had some minor comments, but looks good.

internal/bird/monitor.go
```diff
@@ -110,25 +110,26 @@ func (b *Bird) checkStatus(ctx context.Context) MonitorStatus {
 	cmd := exec.CommandContext(ctx, "birdc", "-s", b.SocketPath, "show", "protocols", "all", `"NBR-*"`)
 	out, err := cmd.CombinedOutput()
 	if err != nil {
-		log.V(1).Info("birdc query failed", "error", err)
+		log.V(1).Info("birdc command failed", "cmd", cmd.String(), "error", err)
```
zolug: Log the CombinedOutput content on birdc failure, as it still discards the actual error message from birdc. Only the command string and exit code are logged, the diagnostic text is lost. It's a generic problem btw.

```go
log.V(1).Info("birdc command failed", "cmd", cmd.String(), "error", err, "output", strings.TrimSpace(string(out)))
```

example:
```
{"level":"info","ts":"2026-08-12T14:38:08Z","logger":"routing","msg":"birdc query failed","error":"exit status 1","output":"Unable to connect to server control socket (/tmp/faulty): No such file or directory"}
```

internal/bird/monitor.go
arp-est: How about if we make this a follow up, of this follow up issue? :'D

arp-est: https://github.com/Nordix/Meridio-2/issues/237

zolug (approved):

Issue #230, opened 2026-08-04
Unvalidated cross-namespace backendRef.namespace in LB/DG controller L34Route mappers


zolug:
## Summary

Both the LoadBalancer controller (`resolveBackendRefDG()`) and the DistributionGroup controller (`extractDGsFromBackendRefs()`) resolve a `gatewayv1.BackendRef` to a `client.ObjectKey`, honoring an explicit `backendRef.Namespace` field with no validation against the caller's scope. The resolved key is then enqueued as a `ctrl.Request` unconditionally in `gatewayEnqueue()` / `l34RouteEnqueue()` (LB controller) and `mapL34RouteToDistributionGroup()` (DG controller) — without checking it matches the controller's expected namespace scope.

No protection against this exists anywhere in the admission/validation path:
- The L34Route CRD schema has no CEL validation on `spec.backendRefs[].namespace` (checked `api/v1alpha1/l34route_types.go`)
- The L34Route validating webhook (`internal/webhook/v1alpha1/l34route_webhook.go`) does not validate `backendRefs` at all — verified by reading the full source: it only checks CIDR/port family consistency and overlap
- No `ReferenceGrant` implementation exists anywhere in the codebase

This is purely a controller-code gap, but the two controllers differ significantly in impact.

---

## LB controller — real bug (destructive)

The LB controller's informer cache is scoped to a single namespace (`cache.Options.DefaultNamespaces: {cfg.GatewayNamespace: {}}`, see `cmd/stateless-load-balancer/cmd/run.go`). A `ctrl.Request` for a different namespace fails `Get` with `NotFound` in `Reconcile`. Critically, `Reconcile` treats `NotFound` as "DistributionGroup deleted" and calls `cleanupDistributionGroup(ctx, req.Name)` — which tears down the NFQLB instance, policy routes, and readiness file, **keyed by name only** (`c.instances[name]`, `c.targets[name]`, `c.flows[name]`), not namespace.

**Impact:** If an L34Route in `c.GatewayNamespace` has a `backendRef` with `kind: DistributionGroup` and an explicit `namespace: <other-ns>` whose **name** happens to collide with a real, actively-managed DG name in `c.GatewayNamespace`, the bogus enqueue triggers a spurious `NotFound` → full cleanup of the real DG's dataplane state. This is a live traffic outage for that DG until the next reconcile re-creates everything.

[...]

PR #235, opened 2026-08-11
Reject cross-namespace backendRef enqueues in LB and DG controllers

bszirtes:
### Issue

#230 

### Problem

Both the LoadBalancer controller (`resolveBackendRefDG()`) and the DistributionGroup controller (`extractDGsFromBackendRefs()`) resolve a `gatewayv1.BackendRef` to a `client.ObjectKey`, honoring an explicit `backendRef.Namespace` with no validation against the caller's scope. The resolved key is enqueued as a `ctrl.Request` unconditionally, without checking it matches the controller's expected namespace scope. No protection exists in the admission path (no CEL validation, no webhook check, no `ReferenceGrant`).

**LB controller impact (destructive):** The LB controller's cache and internal state (`c.instances`/`c.targets`/`c.flows`) are scoped to a single namespace and keyed by name only. A cross-namespace `backendRef` whose name collides with a real DG in the Gateway's namespace triggers a spurious `NotFound` on `Get`, which `Reconcile` treats as "deleted" and tears down the real DG's live NFQLB instance, policy routes, and readiness file — causing a traffic outage until the next reconcile.

**DG controller impact (hygiene):** Same unguarded pattern, currently harmless since cleanup is driven by Kubernetes garbage collection via `ownerReferences`, not by the reconciler's `NotFound` branch. A bogus cross-namespace enqueue just wastes a reconcile cycle today, but matters if cluster-wide operation (`Namespace == ""`) is pursued later.

### Fix

**LB controller** (`internal/controller/loadbalancer/controller.go`):
- Enqueue-time guard in `gatewayEnqueue()` and `l34RouteEnqueue()`: only enqueue a resolved `resolveBackendRefDG()` key if `key.Namespace == c.GatewayNamespace`. `resolveBackendRefDG` itself stays a generic extraction function.
- Defense-in-depth guard at the top of `Reconcile()`: reject any `req.Namespace != c.GatewayNamespace` before the `Get` call, closing off the destructive `NotFound`→cleanup path regardless of which mapper produced the request.
- Scoped the direct-parentRef `DistributionGroupList` call in `gatewayEnqueue()` with `client.InNamespace(c.GatewayNamespace)` for consistency with every other `List` call in the package.

**DG controller** (`internal/controller/distributiongroup/mappers.go`):
- Same enqueue-time guard in `mapL34RouteToDistributionGroup()`, respecting the controller's documented cluster-wide mode (`Namespace == ""` means no restriction applies).

### Out of scope (per issue)
- `belongsToGateway()` — not exploitable today (caller invariant via namespace-scoped cache), no fix requested.
- CEL validation / webhook validation on `backendRefs[].namespace` and a `ReferenceGrant` implementation — noted as context for the gap, not requested deliverables here.
- DG controller cluster-wide mode — explicitly deferred to its own follow-up when that mode is actually pursued.

### Testing
- New/updated tests for every guard, each verified to fail when its corresponding fix is reverted and pass once restored:
  - LB: `Reconcile` namespace guard, `gatewayEnqueue` cross-namespace rejection (both via backendRef and via direct-parentRef DG in another namespace), `l34RouteEnqueue` cross-namespace rejection.
  - DG: `mapL34RouteToDistributionGroup` same-namespace, cross-namespace-rejected, and cluster-wide-passthrough cases.

### DIFF
--- a/internal/controller/loadbalancer/controller_test.go
+++ b/internal/controller/loadbalancer/controller_test.go
@@ -1205,6 +1205,37 @@ var _ = Describe("LoadBalancer Controller", func() {
 			Expect(requests[0].Name).To(Equal("distgroup-1"))
 			Expect(requests[1].Name).To(Equal("distgroup-2"))
 		})
+
+		It("should not enqueue a DG backendRef whose explicit namespace differs from GatewayNamespace", func() {
+			group := meridio2v1alpha1.GroupVersion.Group
+			kind := kindDistributionGroup
+			otherNs := gatewayv1.Namespace("other-ns")
+
+			l34route := &meridio2v1alpha1.L34Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-route",
+					Namespace: namespace,
+				},
+				Spec: meridio2v1alpha1.L34RouteSpec{
+					ParentRefs: []gatewayv1.ParentReference{
+						{Name: gatewayv1.ObjectName(gatewayName)},
+					},
+					BackendRefs: []gatewayv1.BackendRef{
+						{
+							BackendObjectReference: gatewayv1.BackendObjectReference{
+								Group:     (*gatewayv1.Group)(&group),
+								Kind:      (*gatewayv1.Kind)(&kind),
+								Name:      "same-name-dg",
+								Namespace: &otherNs,
+							},
+						},
+					},
+				},
+			}
+
+			requests := controller.l34RouteEnqueue(ctx, l34route)
+			Expect(requests).To(BeEmpty())
+		})
 	})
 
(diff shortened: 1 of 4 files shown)


internal/controller/loadbalancer/controller_test.go
```diff
@@ -1288,6 +1319,42 @@ var _ = Describe("LoadBalancer Controller", func() {
 			Expect(requests[0].Name).To(Equal("direct-dg"))
 		})
 
+		It("should not enqueue a same-named DistributionGroup residing in a different namespace", func() {
```
zolug: desc is misleading: TC looks more like "should not enqueue a DG with direct parentRef when the DG resides in a different namespace"

zolug (approved):

